PR #3, opened 2015-08-12
perf(core) use native modules to compose query strings and MD5 hash

oaleynik:
+ No third-party dependencies
+ Better performance

```
original ix x 38,501 ops/sec ±2.61% (89 runs sampled)
native ix x 56,758 ops/sec ±2.47% (87 runs sampled)
Fastest is native ix
```


kellysutton:
Nice!

Question: some browser-based environments do rely on this plug-in via bower. Is there a way to elegantly handle that, while deferring to crypto/url in the NodeJS/io.js environment?

oaleynik:
@kellysutton yes, I think it is possible. Browserify supports both `crypto` and `url` modules (https://github.com/substack/node-browserify#compatibility). So, this can be compiled to down to browser version without any issues (theoretically).

kellysutton:
Cool. Before I merge this in I'll try to integrate browserify into the build process for the file exposed via the bower interface, `dist/imgix-core-js.umd.js`.

oaleynik:
@kellysutton I used `babelify` as transformation plugin for `browserify` and added this to `compile_umd` npm script. Check this out and let me know if it works for you

kellysutton:
It looks like it works, but it's bloating the UMD output to be 1.5MB. Can you figure out a way to only include `crypto` and  `url`?

It looks like dropping the debug flag will also remove the source map from the end of the file as well, which cuts down the file size.

oaleynik:
@kellysutton `crypto` on its own adds about 570kb after browserification + source map of about 700kb.. Uh, at this point I don't have an idea how to make it efficiently, but I could came up with something later.

oaleynik:
@kellysutton I replaced `browserify` with `webpack`. Webpack does better job - generated UMD file is 10 times smaller

oaleynik:
With `-p` param webpack produces minified file (around 60kb)

kellysutton:
Alright, I think this is definitely moving in the right direction, but there's still some ways to go.

I think deferring to NodeJS for the includes for `crypto` and `url` is a good idea. Increasing the page weight of the bower component will not work, since it's originally 4.7 KB and this change will bring it up to 60 KB. We always prefer to defer to dependency managers when possible and keep our unique bytes of code small.

So with that in mind, would it be possible to do a conditional include? The npm package reads out of the `lib/` directory, which is created from `src/`. The UMD file located in `dist/` will not be used in the NodeJS/io.js environment at all.

Would it be possible to not change (or minimally change) the UMD file, while switching the npm packages over to use `crypto` and `url`?

oaleynik:
@kellysutton some ideas, that came to my head are - use `dependency injection` provided by `webpack http://webpack.github.io/docs/list-of-plugins.html#dependency-injection or use gulp or grunt to cut off chunks of code before transpilation. `DI` works (tried this already), but it requires a bit of restructuring of the codebase, a lot of conditionals (kinda macros) like `if(__BROWSER__) ...` and revisiting of the tests

kellysutton:
@oaleynik Both of these seem like less-than-ideal solutions, as it brings configuration code into the actual code base itself. I think the thing to do here is probably to shim `uri.js` and `md5` into namespaces that map to `crypto` and `url`. This is done in Ember at times to transform globally-style packages in the ES6 namespace. (Here's how you do it in the Ember environment: https://github.com/imgix/ember-cli-imgix-core-js-shim/blob/master/imgix-core-js-shim.js)

What do you think about that?

oaleynik:
@kellysutton I'm totally fine with shimming. `js-md5` module is kind of shim already (more like umd) - it exposes `md5` function on the global context when it is in browser environment and exports module in commonjs style when it runs in node environment as well as falls back to `crypto` module. The bottleneck is `URI-js`. I will have some time on weekends to play with it and figure out how we can fallback to wrap `url` and shim it on browser environment.

kellysutton:
Sounds good. Let me know if I can be of any help!

oaleynik:
@kellysutton I came from another side - in last commit I use `md5` instead of `crypto` (as it falls back to it in commonjs environment) and removed dependency on `url` or `URIjs` by using simplified version of query "stringification."

Cons:
- Doesn't support nested objects or arrays as query string parameters.
 
Pros:
- Performs better
- Doesn't require third-party libs

Also I have an idea of passing `{encode: false}` to `toString` method, which will let me bypass very expensive calls to `encodeURIComponent` and 4-5x times faster URL generation than original implementation.

Let me know what do you think about all of this.

oaleynik:
Side note: it is questionable to use secure key on the client side. Probably, will be better to dedicate this lib just for Node/io.js (which will be the same thing again pretty soon...) or use some kind of public key. Maybe it will be my next PR :)

kellysutton:
@oaleynik We don't recommend using `imgix-core-js` on the client-side when using the secure URL tokens. That removes any security benefits you might gain from secure URL tokens.

Swapping in a custom-built query method is not going to work. We see a lot of issues around hand-rolled URI libraries out there, and we make the best effort to standardize around those. A majority of user errors we see are due to that exact issue.

Let me try to make a shim for URI.js that falls back to `url` in the NodeJS/io.js environment.

kellysutton:
Alright, so I've been doing a bit of research here, and we might be at an impasse. URI.js is the only full featured URL parser/builder that I was able to find that is still being actively developed. We could shim things and defer to `document.createElement('a')` in the browser, but that is not a recommended solution for a few reasons, performance being one of them.

Although we take performance seriously at imgix, this is not something we are willing to optimize at the sake of maintainability. Put another way, we haven't come up with a situation where building imgix URLs should be constrained by performance. (If we are wrong, please let us know.)

That being said, we do need to search for ways of slimming down our bower dependencies to reduce the total payload that we push down to the browser.

I will close this for now, but please reopen if you think we're misguided. I appreciate the effort you put into this, but we think this adds too much complexity for the benefit it provides.

oaleynik:
@kellysutton thanks for letting me know. In my case I need to build hundreds (thousands) of urls almost instantaneously, so performance is essential. For now I will go with my fork in conjunction with aggressive caching strategy.

PR #8, opened 2015-12-10
Replace space in path with the '+' character

### LINKED ISSUE #7: Wrong signature when path contains spaces
Consider the following example:
`http://client-gallery.imgix.net/test/0001 Sample.jpg`

Source tools from the Imgix's webapp (as well as imgix-core-js) generates the following URL:
`http://client-gallery.imgix.net/test/0001 Sample.jpg?s=39a9dc8b6b77986e5208da574ab5f8f2`
... which returns `Unauthorized` response

oaleynik:
Closes #7

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -18,7 +18,7 @@ export class Path {
       this.path = URI.encode(this.path);
     } else {
       // Replace spaces in path with '+'
-      this.path = this.path.replace(/ /g, "+");
+      this.path = URI.encodeReserved(this.path);
     }
 
     if (this.path[0] !== "/") {
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -16,6 +16,9 @@ export class Path {
     // We are dealing with a fully-qualified URL as a path, encode it
     if (this.path.indexOf("http") === 0) {
       this.path = URI.encode(this.path);
+    } else {
+      // Replace spaces in path with '+'
+      this.path = this.path.replace(/ /g, "+");
     }
 
     if (this.path[0] !== "/") {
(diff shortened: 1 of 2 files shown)


oaleynik:
@kellysutton I'm debating with myself what is better `+` or `%20` in this case. I used `+` as S3 console does it.

kellysutton:
Not quite sure what will be best. Our newest hire, @paulstraw, can help you get to the bottom of things. Thanks for the PR!

paulstraw:
I just did some digging on this. [This blog post](http://blog.lunatech.com/2009/02/03/what-every-web-developer-must-know-about-url-encoding#CommonpitfallsofURLs) has a great overview, but basically it looks like since this is a path and not a query, the ` ` characters should be encoded as `%20`. Let's go with that for now, if you don't mind changing it.

After this is merged in, I'm actually planning to more "fully" sanitize passed paths according to the "unsafe" section of [IETF RFC 1738](http://www.ietf.org/rfc/rfc1738.txt) (since we unfortunately can't just `URI.encode` them). I'll be digging into how to best make that happen in #9.

paulstraw:
@oaleynik, it looks like `URI.encodeReserved` would be a great fit here. If you want to make that change, I'll be happy to merge it in!

oaleynik:
@paulstraw thanks for your comment. Yeah, I knew that direct string manipulation is not perfect. Your suggestion to use `URI.encodeReserved` works perfectly. Thanks!

oaleynik:
As a side note - I saw that post related to URL encoding when I was digging on this issue. I don't think that the `+` is wrong entirely or against IEFT specs. Also, it looks better in the URL:

https://my-social-network.imgix.net/users/image%201.png?s=193462f12470fe53927d0cf21e07d404
https://my-social-network.imgix.net/users/image+1.png?s=193462f12470fe53927d0cf21e07d404

Anyway - I subscribed to #9 to follow your findings. Thanks again!

PR #16, opened 2016-03-01
ES5 Rewrite

paulstraw:
@kellysutton, I'd appreciate some eyes on this when you've got some time.


src/imgix-core-js.js
```diff
@@ -1,103 +1,126 @@
...
+        params.ixlib = this.settings.libraryParam
+      }
 
-    return query;
-  }
+      var queryParams = [];
+      var key, val, encodedKey, encodedVal;
+      for (key in params) {
+        val = params[key];
+        encodedKey = encodeURIComponent(key);
```
kellysutton: What's the scenario where we need to encode our own keys? Seems a little weird.

src/imgix-core-js.js
```diff
@@ -1,103 +1,126 @@
...
-  toUrl(newParams) {
-    this.queryParams = Object.assign(this.queryParams, newParams);
-    return this;
-  }
+      if (!this.settings.host) {
+        throw new Error('ImgixClient must be passed a valid host');
+      }
 
-  _query() {
-    return URI.buildQuery(Object.assign(this._queryWithoutSignature(), this._signature()), false, false);
```
kellysutton: I'm not crazy about removing URI.js, since I always consider building URLs "by hand" to be a bit of a smell. Given that this eliminates a third-party dependency, I think it's okay.

src/imgix-core-js.js
paulstraw: This is a new standard across all libraries. It's an unlikely case, but protects against anyone generating URLs from user input.

src/imgix-core-js.js
```diff
@@ -1,103 +1,126 @@
...
-  }
+      if (this.settings.includeLibraryParam) {
+        this.settings.libraryParam = "js-" + VERSION;
+      }
+
+      if (this.settings.useHTTPS) {
+        this.settings.urlPrefix = 'https://'
```
kellysutton: Good case for a ternary operator here.

src/imgix-core-js.js
```diff
@@ -1,103 +1,126 @@
...
+
+      if (params == null) {
+        params = {};
+      }
 
-  _b64EncodedQuery() {
-    let query = this.queryParams;
-    let encodedQuery = {};
+      var queryParams = this._buildParams(params);
+      if (this.settings.secureURLToken != null) {
```
kellysutton: Would this be better as a `!!`? That way, it will also fail if you pass the empty string.

kellysutton:
Left a few notes, nothing major. Review and :shipit:

src/imgix-core-js.js
kellysutton: :+1:

src/imgix-core-js.js
paulstraw: I 100% agree with this in almost every case. However, with the imgix libraries, I'd like to try to avoid third-party URL building libraries, since the imgix libraries themselves are effectively glorified URL building libraries.

paulstraw:
Just gotta finish updating the readme and I'll get this shipped. Then Sandbox can finally be updated!

PR #22, opened 2016-11-30
Update js-md5 to v0.4

NGMarmaduke:
When running in a certain server runtimes v0.3.0 makes the assertion
that the `navigator` option is available on the global scope resulting
in the error `Cannot read property 'navigator' of undefined`.

This appears to be fixed in v0.4.

The offending line can be found here: https://github.com/emn178/js-md5/blob/v0.3.0/src/md5.js#L20

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "main": "dist/imgix-core-js.js",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


paulstraw:
Thanks @NGMarmaduke! This has been released as 1.0.4.

Issue #24, opened 2017-04-06
Incorrect url encoding


MichaelMag:
Origin file name: "download copy #1-[mike].png"
S3 URL: https://s3-us-west-2.amazonaws.com/testBuck/download%20copy%20%231-%5Bmike%5D.png

`ImgixClient.buildURL(fileName, params)` should return:  "https://some-prefix.imgix.net/download%20copy%20%231-%5Bmike%5D.png?s=52dd73573bffb0948d49118f30ea9218
but the result is: https://some-prefix.imgix.net/download%20copy%20#1-%5Bmike%5D.png?s=12c9ae700d245a639334a2763c93c304

So, the result url is incorrect because of `#` in the url and imgix return _URI signature match failed_.
I found that the problem is here: 
https://github.com/imgix/imgix-core-js/blob/1.0.5/src/imgix-core-js.js#L65-L83

jayeb:
@MichaelMag Sorry about the delay here--up until today I wasn't actually watching this repo like I thought I was 😞 . Will take a look this week.

jayeb:
Additional, related issue: if a URL is passed in "partially" encoded (for example, if there's a query parameter with an encoded value, but the URL is otherwise unencoded), the decode step on L70 is harming the input URL.

jayeb:
Just released 1.0.6, which fixes this issue.

PR #25, opened 2017-05-24
Rework de/encoding in `_sanitizePath` method to be more careful

jayeb:
This PR is intended to fix the two issues brought up in #24, both of which involve encoding errors brought about by ham-fistedly decoding all paths fed in to the `_sanitizePath()` private method before attempting to re-encode things properly based on whether we determined the path was a fully qualified URL or just a path fragment.

Instead of this clumsy approach, we're now evaluating the given path up front to determine which of the following categories it falls into. Each category is then handled slightly differently:

- *Fully-qualified, unencoded URLs*. If the given path matches `/^https?:\/\//`, we assume its a fully-qualified, unencoded URL. All we need to do here is encode it as though it was a URI component (since it will be just a component, in the final imgix URL).
- *Fully-qualified, pre-encoded URLs*. If the given path matches `/^https?:%3A%2F%2F/`, we assume its a fully-qualified, pre-encoded URL. In this case, we decode and re-encode it using `decodeURIComponent` and `encodeURIComponent`, to ensure that all characters in the string are actually encoded properly.
- *Path fragments*. If the given path does not match either of the above patterns, we must assume that it's a boring path rather than a full URL. In this case, we decode and re-encode it as above, but this time we use `decodeURI` and `encodeURI` so that legal path characters such as `/` and `@` aren't affected.

This PR also reorganizes the library's tests so they're focused on the private component methods that make up the public `buildURL()` method: `_sanitizePath()`, `_buildParams()`, and `_signParams()`. Testing these component functions separately allows us to untangle the various cases that need to be tested and gives us better confidence that all cases are being covered.

@mchatman Do you mind giving this a quick review?

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -66,17 +66,19 @@
       // Strip leading slash first (we'll re-add after encoding)
       path = path.replace(/^\//, '');
 
-      // Decode the path, since we don't know whether it's pre-encoded
-      path = decodeURIComponent(path);
-
       if (/^https?:\/\//.test(path)) {
-        // Use encodeURIComponent if we think the path is a full URL,
-        // since it encodes all characters
+        // Use de/encodeURIComponent to ensure *all* characters are handled,
+        // since it's being used as a path
         path = encodeURIComponent(path);
+      } else if (/^https?%3A%2F%2F/.test(path)) {
+        // Decode and re-encode the path, to ensure that it's fully, correctly encoded.
+        // Use de/encodeURIComponent to ensure *all* characters are handled,
+        // since it's being used as a path
+        path = decodeURIComponent(encodeURIComponent(path));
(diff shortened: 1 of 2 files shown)


mchatman:
Looks great. Worked successfully with an URL that previously failed.

mixergtz:
Hi @jayeb, thanks for the awesome job!

I'm wondering if there's some way to skip the double encoding you added in this PR, my use case is when you have an already encoded url like `https://mysite.s3.amazonaws.com/My%20Product.jpg` you'll get an extra-encoded url looking like `https://mysite.s3.amazonaws.com/My%2520Product.jpg` which won't be present in the server.
My current workaround is to decode the url before passing it to `buildUrl` I'll be thankful to hear any thoughts from you on this matter.

PR #26, opened 2017-07-13
Use standalone md5 package

joshfrench:
The `js-md5` package requires `crypto` under the hood. Webpack 2+ "helpfully" shims this in the browser by bringing in https://www.npmjs.com/package/node-libs-browser. This can add ~100kb to the bundled output.

The `md5` package has no such dependencies and is a drop-in replacement. As a result, Webpack and other module bundlers should see significant size reductions to projects that depend on `imgix-core-js`.


jayeb:
This looks great, thank you @joshfrench. For organization's sake, can you update this commit to only include the changes to `src/imgix-core-js.js` and `package.json`? We prefer to only build and check in generated files on master at the same time we bump version numbers.

joshfrench:
@jayeb All set!

PR #27, opened 2017-07-26
generate img tags and srcsets

selfagency:
i created methods for generating image tags and srcsets instead of just plain old URLs


jayeb:
Thanks @selfagency!

Adding new features like this is something we're very careful about. We're keenly aware that libraries that take a "kitchen sink" approach to adding features tend to be very hard to maintain, so we make sure that each new feature we add is carefully considered and adds value to the library for a large number of users.

Can you describe to me your use-case for these functions? Are you using this code on the front-end in conjunction with some DOM manipulation, or are you using this as part of a templating system in a Node environment? Or something else entirely?

selfagency:
I'm using this in conjunction with Handlebars generated templates in Node. Because I use secure sources and sign my images, I can't use imgix.js to create srcsets on the frontend. Therefore, after one of your support techs told me effectively I would likely be waiting quite a while for this ability, I opted to demonstrate how this could be done on the backend readily.

jayeb:
Hi @selfagency,

After some discussion internally, we've decided not to merge this PR as-is. We feel pretty strongly that having HTML-generating functions in this library is a bad path for us to go down--in order for us to do it right, we'd need to add a lot more complexity to this codebase than we're comfortable doing (handling `<picture>` tags, arbitrary attributes, proper sanitization, etc.)

However, we think you're right about the usefulness of having a method on the `ImgixClient` that helps in generating a `srcset` list. If done correctly, a function that spits out a map or string of optionally-signed URLs over a range of widths would be broadly useful to a lot of users. We'll be putting together a PR for this soon.

PR #29, opened 2017-08-25
Remove "clever" decoding from `_sanitizePath`

### LINKED ISSUE #28: If filename in path has spaces - signature becomes invalid
Consider the following example:

```javascript
var ImgixClient = require('imgix-core-js');

  host: "example.com",
  secureURLToken: "12345abcde",
  secure: true
});

var url = client.buildURL("/a/b/c/image with spaces.jpg", {
  fm: 'jpg',
  dpr: 2,
  w: 1500,
  sharp: 15
});
```

The above outputs: `https://example.com/a/b/c/image with spaces.jpg?fm=jpg&dpr=2&w=1500&sharp=15&ixlib=js-1.0.7&s=5d03453c163fb3c22722a1b7d3fc874a` when tool from web app outputs spaces URL encoded.

Request to the resulting URL returns 403 Invalid Signature. I needed to pin imgix-core-js version to 1.0.5 to avoid this.

jayeb:
I had previously updated the `ImgixClient._sanitizePath()` method in an attempt to be clever and properly handle pre-encoded paths. This introduced unnecessary complexity into the method (detemining whether or not a given string is unencoded, sorta pre-encoded, or fully pre-encoded is trickier than it seems at first glance) and introduced stupid bugs into its base use-case.

Having failed my attempt at cleverness, I am suitably chagrined. This PR removes all code dealing with pre-encoded paths from `ImgixClient._sanitizePath()` so it only handle non-encoded inputs. Our README has instructed users to only pass in non-encoded data from the start:

> Whenever you provide data to imgix-core-js, make sure it is not already URL-encoded, as the library handles proper encoding internally.

This also fixes #28, and includes a new test case to ensure non-URL paths with spaces are handled properly. 

@oaleynik Would you care to review these changes for me?

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -70,15 +70,10 @@
         // Use de/encodeURIComponent to ensure *all* characters are handled,
         // since it's being used as a path
         path = encodeURIComponent(path);
-      } else if (/^https?%3A%2F%2F/.test(path)) {
-        // Decode and re-encode the path, to ensure that it's fully, correctly encoded.
-        // Use de/encodeURIComponent to ensure *all* characters are handled,
-        // since it's being used as a path
-        path = decodeURIComponent(encodeURIComponent(path));
       } else {
         // Use de/encodeURI if we think the path is just a path,
         // so it leaves legal characters like '/' and '@' alone
-        path = decodeURI(encodeURI(path));
+        path = encodeURI(path);
       }
 
       return '/' + path;
--- a/test/test-client.js
+++ b/test/test-client.js
@@ -77,8 +77,8 @@ describe('Imgix client:', function describeSuite() {
       });
     });
 
-    describe('with a path that includes encoded characters', function describeSuite() {
-      var path = 'images/image%201.png';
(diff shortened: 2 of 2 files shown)


oaleynik:
@jayeb thanks for fixing that! I guess the problem we had previously in #7 was due to the use of encodeURIComponent for the path part instead of encodeURI. encodeURI should be handling reserved characters appropriately (per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI), so I think this is good to go.

jayeb:
I'll be releasing this as 1.1. While this is sort of a patch-level change in that it's fixing a bug, it is technically changing the interface of `ImgixClient._sanitizePath()`. Even though it's changing the interface to more closely match the documented behavior, I'd rather push this as a minor-level change just in case someone somewhere had an implementation that was dependent on the non-documented behavior of that method.

oaleynik:
@jayeb that makes perfect sense. Thanks again for fixing this quickly! 🤘🏻

PR #30, opened 2017-10-31
Fixed Base64 issue when define is defined

nkishorchandra:
Fixed an issue with Base64 encoding when ImgixClient is initialized from the first case, i.e.
if (typeof define === 'function' && define.amd) {
    define('Imgix', ['exports', 'md5', 'js-base64'], factory);

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -11,7 +11,7 @@
   }
 })(this, function (exports, _md5, _jsBase64) {
   var md5 = _md5;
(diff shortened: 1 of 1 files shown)


jayeb:
Good catch, @nkishorchandra. Can you amend your commit to *only* update the `src/imgix-core-js.js` file? I'll build and version the dist files once this is merged into master.

nkishorchandra:
Done @jayeb. Please check.

jayeb:
Looks great, thank you!

PR #34, opened 2018-06-21
Support CRC and Cyclic Domain Sharding

hashknot:


test/test-client.js
```diff
@@ -290,4 +318,76 @@ describe('Imgix client:', function describeSuite() {
       assert.equal(expectation, result);
     });
   });
+
+  describe('Sharding', function describeSuite() {
+    describe('CRC', function describeSuite() {
+      it('path resolves to same domain', function testSpec() {
+        var domains = ['my-host1.imgix.net', 'my-host2.imgix.net']
```
jayeb: [nitpick] Missing semicolon.

test/test-client.js
```diff
@@ -290,4 +318,76 @@ describe('Imgix client:', function describeSuite() {
...
+        assert.ok(client.buildURL('/users/a.png').match(domains[0]));
+        assert.ok(client.buildURL('/users/2.png').match(domains[0]));
+        assert.ok(client.buildURL('/users/b.png').match(domains[0]));
+        assert.ok(client.buildURL('/users/a.png').match(domains[0]));
+      });
+    });
+
+    describe('Cyclic', function describeSuite() {
+      it('domains cycle', function testSpec() {
+        var domains = ['my-host1.imgix.net', 'my-host2.imgix.net', 'my-host3.imgix.net']
```
jayeb: [nitpick] Missing semicolon.

test/test-client.js
```diff
@@ -290,4 +318,76 @@ describe('Imgix client:', function describeSuite() {
...
+        assert.ok(client.buildURL('/users/a.png').match(domains[1]));
+      });
+
+      it('single domain sharding', function testSpec() {
+        var domains = 'my-host1.imgix.net';
+        var client = new ImgixClient({
+          domains: domains,
+          shard_strategy: ImgixClient.SHARD_STRATEGY_CRC
+        });
+        assert.ok(client.buildURL('/users/1.png').match(domains[0]));
```
jayeb: I don't think these tests are actually doing the thing you want them to. Since `domains` is a string here and not an array, this is just checking whether the first character of the string appears in the resultant URL.

As a sidenote, if we could rename this variable to `domain` in this test, that would be good.

src/imgix-core-js.js
```diff
@@ -59,9 +81,18 @@
...
+      return this.settings.urlPrefix + this._getDomain(path) + path + queryParams;
     };
 
+    ImgixClient.prototype._getDomain = function(path) {
+      if(this.settings.shard_strategy === SHARD_STRATEGY_CYCLE) {
+        var domain = this.settings.domains[this._shard_next_index];
+        this._shard_next_index = (this._shard_next_index + 1) % this.settings.domains.length;
+        return domain;
+      }
+      return this.settings.domains[crc.crc32(path) % this.settings.domains.length];
```
jayeb: Can we wrap this line in an `else if (this.settings.shard_strategy === SHARD_STRATEGY_CRC) { ... }`? I'd rather be verbose than obscure.

src/imgix-core-js.js
```diff
@@ -117,6 +148,8 @@
     };
 
     ImgixClient.VERSION = VERSION;
+    ImgixClient.SHARD_STRATEGY_CRC = SHARD_STRATEGY_CRC;
```
jayeb: I don't really like exposing these constants--it makes things a little bit easier in our unit tests, but kind of muddies up the `ImgixClient` object. Can we remove these two lines and just redefine these constants somewhere in the test harness instead?

src/imgix-core-js.js
```diff
@@ -36,8 +42,24 @@
...
+      }
+
+      if (this.settings.shard_strategy !== SHARD_STRATEGY_CRC
+          && this.settings.shard_strategy !== SHARD_STRATEGY_CYCLE) {
+        throw new Error('Shard strategy must be one of ' +
+          SHARD_STRATEGY_CRC + ', ' + SHARD_STRATEGY_CYCLE);
+      }
+
+      if(this.settings.host){
+        // console.warn("'host' argument is deprecated; use 'domains' instead.")
```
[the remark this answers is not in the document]
hashknot: I left it commented by mistake.

I didn't use `console.error` because `host` is still a valid (but deprecated) argument. Browser consoles highlight `console.warn` and `console.warn` messages but not `console.log`.

src/imgix-core-js.js
hashknot: The constants aren't just for testing. We expect library users to specify the shard strategy while instantiating `ImgixClient`.

test/test-client.js
hashknot: Makes sense. Will update it.

src/imgix-core-js.js
```diff
@@ -36,8 +42,24 @@
         this.settings[key] = val;
       }
 
-      if (!this.settings.host) {
-        throw new Error('ImgixClient must be passed a valid host');
+      if(!Array.isArray(this.settings.domains)){
```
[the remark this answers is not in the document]
hashknot: Got it.

src/imgix-core-js.js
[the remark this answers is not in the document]
jayeb: Yea, browsers handle it nicely but in Node `console.warn` [is an alias for `console.error`](https://nodejs.org/api/console.html#console_console_warn_data_args). Given that this prints to `stderr`, I think it's probably safer to avoid it and use `console.log`.

src/imgix-core-js.js
jayeb: Alright, fair point.

src/imgix-core-js.js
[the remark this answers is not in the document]
hashknot: Isn't the warning messages being redirected to `stderr` a good thing? You can easily redirect `stderr` to a file or `/dev/null` without polluting the library-importing-code's `console.log` output messages.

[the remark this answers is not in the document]
jayeb: Yea, you're right. Best to leave it as-is.

PR #37, opened 2019-02-27
Add domain validation during ImgixClient initialization

sherwinski:
Adds a validation that will throw an error if a user attempts to initialize a new `ImgixClient` object with an invalid domain.
The validation criteria is based off of the same regex used on [dashboard.imgix.com](https://dashboard.imgix.com)

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -1,3 +1,5 @@
+var constants = require("../src/constants.js");
+
 (function (global, factory) {
   if (typeof define === 'function' && define.amd) {
     define('Imgix', ['exports', 'md5', 'js-base64', 'crc'], factory);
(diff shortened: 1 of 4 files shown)


src/imgix-core-js.js
```diff
@@ -62,8 +61,17 @@
           this.settings.domains[0] = this.settings.host;
       }
 
+      this.settings.domains.forEach(function(domain) {
+        if(constants.DOMAIN_REGEX.exec(domain) == null){
```
jayeb: [nit] Codestyle here dictates there should be spaces around the parens: `if (...) {`

src/imgix-core-js.js
```diff
@@ -62,8 +61,17 @@
...
+          throw new Error(
```
jayeb: [nit] Wouldn't mind seeing this re-indented:

```js
throw new Error(
  'Domains must be passed in as fully-qualified ' + 
  'domain names and should not include a protocol or any path ' + 
  'element, i.e. "example.imgix.net".'
);
```

PR #42, opened 2019-06-01
fix: remove deprecated domain sharding functionality

sherwinski:
This PR removes all domain sharding functionality from imgix-core-js, following its deprecation in #39


src/imgix-core-js.js
```diff
@@ -1,37 +1,32 @@
...
-  var crc = _crc;
 
   var VERSION = '1.3.0';
-  var SHARD_STRATEGY_CRC = 'crc';
-  var SHARD_STRATEGY_CYCLE = 'cycle';
   var DOMAIN_REGEX = /^(?:[a-z\d\-_]{1,62}\.){0,125}(?:[a-z\d](?:\-(?=\-*[a-z\d])|[a-z]|\d){0,62}\.)[a-z\d]{1,63}$/i;
   var DEFAULTS = {
     host: null,
-    domains: [],
+    domain: '',
```
sherwinski: One thing that I'm a little iffy on: we mention in the last release that domain sharding would be deprecated but failed to mention this change to the `domains` field specifically. Unlike in [imgix-rb](https://github.com/imgix/imgix-rb/blob/be8b5d55517b107aa9373c964b6049855e6f842d/lib/imgix/client.rb#L17), we only accept a `domains` argument, which means that users who don't use sharding would still be affected by this release. Because of that, I wonder if this change would still be considered kosher.

A compromise could be deprecating (but still respecting) the `domains` field, while throwing a warning. If that's the case, I'm thinking might be cleaner to just pull that change out into a separate deprecation release before moving forward with this PR.

jayeb: I agree with the "compromise" approach. Releases are free, so we shouldn't be afraid of making as many releases as necessary to ensure a smooth ramp for our users.

test/test-client.js
```diff
@@ -1,112 +1,58 @@
 var assert = require('assert');
 var ImgixClient = require('../src/imgix-core-js');
-var sinon = require('sinon');
```
jayeb: Is there any reason not to just leave sinon in here for use in the future? It's listed as a dev dependency, so keeping it around won't end up affecting any end-users.

sherwinski: Good point, I did it for cleanliness but I'm not opposed to keeping it - especially if it's needed again in the future.

src/imgix-core-js.js
```diff
@@ -1,38 +1,32 @@
...
 
   var VERSION = '1.4.0';
-  var SHARD_STRATEGY_CRC = 'crc';
-  var SHARD_STRATEGY_CYCLE = 'cycle';
   var DOMAIN_REGEX = /^(?:[a-z\d\-_]{1,62}\.){0,125}(?:[a-z\d](?:\-(?=\-*[a-z\d])|[a-z]|\d){0,62}\.)[a-z\d]{1,63}$/i;
   var DEFAULTS = {
     host: null,
-    domain: null,
-    domains: [],
+    domain: '',
```
jayeb: Can we continue using `null` as the default? Seems like it would be easier to catch when the option is omitted if it defaults to `null`.

src/imgix-core-js.js
```diff
@@ -44,47 +38,25 @@
...
-            throw new Error('ImgixClient.settings.domain only accepts a string argument');
-          }
-          else {
-            this.settings.domains = [this.settings.domain];
-          }
-        }
-      }
-      else {
-        this.settings.domains = [this.settings.domains];
+      if (typeof(this.settings.domain) != "string") {
```
jayeb: Does this work? This should not work. Should be:

```js
if (typeof this.settings.domain != "string") {
```

src/imgix-core-js.js
```diff
@@ -44,47 +38,25 @@
...
+        throw new Error('ImgixClient must be passed valid domain');
       }
 
       if (this.settings.host) {
-        console.warn("'host' argument is deprecated; either use 'domain' or 'domains' instead.");
-        if (this.settings.domains.length == 0)
-          this.settings.domains[0] = this.settings.host;
+        console.warn("'host' argument is deprecated; use 'domain' instead.");
+        if (this.settings.domain.length == 0)
+          this.settings.domain = this.settings.host;
```
jayeb: Should we go ahead and remove `host` as an option too, since we're already bumping the major version here?

src/imgix-core-js.js
jayeb: Also, I'd suggest we just combine this with the second check, and just use "ImgixClient must be passed valid string domain" as the error message.

src/imgix-core-js.js
sherwinski: Ah yea, good catch

src/imgix-core-js.js
sherwinski: Yea I was planning to do that in a separate PR

src/imgix-core-js.js
sherwinski: I think so? It's what we did [last time](https://github.com/imgix/imgix-core-js/blob/ca889dc09dd153dfca889d7aaa6da2fde1cf7c5b/src/imgix-core-js.js#L52) but I can double check

jayeb: Oh god, I should have caught that. `typeof` is an operator, not a function. That needs to be fixed.

PR #43, opened 2019-06-03
docs: deprecate settings.domains

sherwinski:
This PR adds further detail to the sharding deprecation warning, alerting users that `settings.domains` will be deprecated in favor of `settings.domain`. This change should precede the domain sharding removal release.

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -47,7 +47,7 @@
         this.settings.domains = [this.settings.domains];
       }
       else {
-        console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.");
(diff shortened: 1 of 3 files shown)


src/imgix-core-js.js
```diff
@@ -47,7 +47,7 @@
         this.settings.domains = [this.settings.domains];
       }
       else {
-        console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.");
+        console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.\nAs a result, the 'domains' argument will be deprecated in favor of 'domain' instead.");
```
sherwinski: @jayeb I tagged you for review in case you think there's a better way to communicate this

jayeb: Naw, seems fine.

PR #44, opened 2019-06-04
feat: add settings.domain argument

sherwinski:
This PR is a follow up to #43, creating a new field in the `ImgixClient` object that will replace `settings.domains` once domain sharding is removed.

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -78,7 +78,7 @@
       }
 
       if (this.settings.host) {
-        console.warn("'host' argument is deprecated; use 'domains' instead.");
+        console.warn("'host' argument is deprecated; either use 'domain' or 'domains' instead.");
         if (this.settings.domains.length == 0)
           this.settings.domains[0] = this.settings.host;
       }
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -44,33 +44,27 @@
         this.settings[key] = val;
       }
 
-      // If settings.domains is passed in as an array of 1 or more elements
-      // will generate a deprecation warning
-      if (Array.isArray(this.settings.domains) && this.settings.domains.length > 0) {
-        console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.\nAs a result, the 'domains' argument will be deprecated in favor of 'domain' instead.");
-      }
-      // If settings.domains is passed in as a signle domain str
-      // will keep original behavior
-      else if (!Array.isArray(this.settings.domains)) {
-        this.settings.domains = [this.settings.domains];
-      }
-      // If new domain argument is being used
-      else {
-        // If domain is passed an array
-        // check if it is an array > 1 elements
-        if (Array.isArray(this.settings.domain)) {
-          if(this.settings.domain.length > 1) {
-            throw new Error('ImgixClient.settings.domain cannot take multiple domains');
+      if (Array.isArray(this.settings.domains)) { 
+        if (this.settings.domains.length > 1) {
+          console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.\nAs a result, the 'domains' argument will be deprecated in favor of 'domain' instead.");
+        }
+        else if (this.settings.domains.length != 1)
+        {
+          if (Array.isArray(this.settings.domain)) {
+            if(this.settings.domain.length > 1) {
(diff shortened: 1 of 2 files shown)


src/imgix-core-js.js
```diff
@@ -43,15 +44,37 @@
         this.settings[key] = val;
       }
 
-      if (!Array.isArray(this.settings.domains)) {
+      // If settings.domains is passed in as an array of 1 or more elements
+      // will generate a deprecation warning
+      if (Array.isArray(this.settings.domains) && this.settings.domains.length > 0) {
```
jayeb: Don't we want to only raise this warning if `this.settings.domains.length > 1`? By definition, someone can't be using sharding if they're only providing one domain.

src/imgix-core-js.js
```diff
@@ -43,15 +44,37 @@
...
+      // will keep original behavior
+      else if (!Array.isArray(this.settings.domains)) {
         this.settings.domains = [this.settings.domains];
       }
+      // If new domain argument is being used
       else {
-        console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.\nAs a result, the 'domains' argument will be deprecated in favor of 'domain' instead.");
+        // If domain is passed an array
+        // check if it is an array > 1 elements
+        if (Array.isArray(this.settings.domain)) {
```
jayeb: I don't think we should even allow `domain` to be passed as an array.

sherwinski: Agreed, I figured this would be a way to mitigate it if users tried to. I'm open to other ways that accomplish the same thing.

src/imgix-core-js.js
sherwinski: That was more of a consequence of L24 where domains is defaulted to `[]`. But I agree with your point especially because I do the same thing a few more lines down.

sherwinski:
The way I was looking at it, `domains` is still the default as these changes would be part of a minor release, preceding sharding removal. I do think we should switch over in the new release once we completely get rid of `domains`. Maybe this approach is less clear, but I think framing it that way informed why I went this approach.

jayeb:
Yea okay, that's a fair point.

src/imgix-core-js.js
```diff
@@ -43,15 +44,26 @@
...
+        }
       }
       else {
-        console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.\nAs a result, the 'domains' argument will be deprecated in favor of 'domain' instead.");
+        this.settings.domains = [this.settings.domains];
       }
-
-      if (!this.settings.host && this.settings.domains.length === 0) {
-        throw new Error('ImgixClient must be passed valid domain(s)');
+
```
jayeb: [nit] Whitespace

src/imgix-core-js.js
```diff
@@ -43,15 +44,26 @@
...
       }
 
-      if (!Array.isArray(this.settings.domains)) {
-        this.settings.domains = [this.settings.domains];
+      if (Array.isArray(this.settings.domains)) { 
+        if (this.settings.domains.length > 1) {
+          console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.\nAs a result, the 'domains' argument will be deprecated in favor of 'domain' instead.");
+        }
+        else if (this.settings.domains.length == 0)
+        {
```
jayeb: [nit] This curly should be on the previous line.

src/imgix-core-js.js
```diff
@@ -43,15 +44,26 @@
...
       else {
-        console.warn("Warning: Domain sharding has been deprecated and will be removed in the next major version.\nAs a result, the 'domains' argument will be deprecated in favor of 'domain' instead.");
+        this.settings.domains = [this.settings.domains];
       }
-
-      if (!this.settings.host && this.settings.domains.length === 0) {
-        throw new Error('ImgixClient must be passed valid domain(s)');
+      
+      if (!this.settings.host && this.settings.domains == 0) {
+          throw new Error('ImgixClient must be passed valid domain(s)');
```
jayeb: [nit] Indentation

src/imgix-core-js.js
sherwinski: I think that's fine since it's separating two different if/else blocks?

jayeb: The newline is fine, but the line should be empty.

PR #51, opened 2019-07-10
ci(travis): update node versions

sherwinski:
This PR removes node v4/adds v12 in the travis config
[More info](https://nodejs.org/en/about/releases/) on node LTS

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
   "license": "BSD-2-Clause",
   "devDependencies": {
     "mocha": "2.5.3",
-    "phantomjs": "1.9.20",
     "sinon": "7.3.2",
     "uglify-js": "3.6.0"
   }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -17,7 +17,6 @@
   "license": "BSD-2-Clause",
   "devDependencies": {
     "mocha": "2.5.3",
-    "phantomjs": "1.9.20",
```
jayeb: What's the story here?

sherwinski: I just wanted to test it out and see if anything else broke. Do we actually use this anywhere?

jayeb: It's almost certainly used by Mocha when running unit tests, but if things aren't breaking perhaps it's included in there as a peer dependency?

sherwinski: How about a compromise: we keep phantomjs but upgrade to prevent strange testing failures?

jayeb: It's a deal 🤝.

PR #53, opened 2019-07-18
feat: add srcset generation

sherwinski:
This PR adds a new function to the imgix-core-js interface that will either generate a width descriptor or DPR `srcset` based on the specified parameters.
Special attention should be given to cover all aspect ratio-related cases, especially with the newly available `ar` parameter.

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -71,44 +71,28 @@
     };
 
     ImgixClient.prototype.buildSrcSet = function (path, params) {
-      var width = params['w'];
-      var height = params['h'];
-      var aspectRatio = params['ar'];
-      var aspectRatioAsDecimal = function() {
-        if (typeof aspectRatio !== "string") {
-          return false;
-        }
-        var isValidFormat = function(str) {
-          return /^\d+(\.\d+)?:\d+(\.\d+)?$/.test(str);
-        };
-        if (!isValidFormat(aspectRatio)) {
-          return false;
-        }
       
-        const aspectRatioSplit = aspectRatio.split(":");
-        var arWidth = aspectRatioSplit[0];
-        var arHeight = aspectRatioSplit[1];
-      
-        return parseFloat(arWidth) / parseFloat(arHeight);
+      var width = params ? params['w'] : undefined;
+      var height = params ? params['h'] : undefined;
+      var aspectRatio = params ? params['ar'] : undefined;
+
+      // determines if an aspect ratio value is in the correct format 'w:h'
+      var isValidFormat = function() {
+        return /^\d+(\.\d+)?:\d+(\.\d+)?$/.test(aspectRatio);
       }();
+
+      if (aspectRatio && !isValidFormat) {
+        throw new Error('The \'ar\' parameter key must follow the format w:h');
+      }
 
       var fixedWidth = ((width && height) || (width && aspectRatio) || (height && aspectRatio)) ? true : false;
 
       if (fixedWidth) {
-        if (width && height) {
-          return this._buildDPRSrcSet(path, params);
-        }
-        else if (aspectRatio) { 
-          if (width) {
-            params['h'] = width / aspectRatioAsDecimal;
-          }
-          else if (height) {
-            params['w'] = height * aspectRatioAsDecimal;
-          }
-          return this._buildDPRSrcSet(path, params);
-        }
+        return this._buildDPRSrcSet(path, params);
       }
-      return this._buildSrcSetPairs(path, params);
+      else {
+        return this._buildSrcSetPairs(path, params);
+      }
     };
 
     ImgixClient.prototype._buildSrcSetPairs = function(path, params) {
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -150,16 +150,6 @@
       var width = params ? params['w'] : undefined;
       var height = params ? params['h'] : undefined;
       var aspectRatio = params ? params['ar'] : undefined;
-      if (params) {
-        delete params.dpr;
-      }
-      
-      // determines if an aspect ratio value is in the correct format 'w:h'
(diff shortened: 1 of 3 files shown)


src/imgix-core-js.js
```diff
@@ -70,6 +70,72 @@
...
+        return parseFloat(arWidth) / parseFloat(arHeight);
+      }();
+
+      var fixedWidth = ((width && height) || (width && aspectRatio) || (height && aspectRatio)) ? true : false;
+
+      if (fixedWidth) {
+        if (width && height) {
+          return this._buildDPRSrcSet(path, params);
+        }
+        else if (aspectRatio) {
```
sherwinski: At first I thought that these aspect ratio calculations were necessary but the more I mess around with the `ar` parameter, the less I think this is actually needed.

src/imgix-core-js.js
```diff
@@ -124,6 +124,74 @@
       }
     };
 
+    ImgixClient.prototype.buildSrcSet = function (path, params) {
+      var width = params ? params['w'] : undefined;
+      var height = params ? params['h'] : undefined;
+      var aspectRatio = params ? params['ar'] : undefined;
+
+      // determines if an aspect ratio value is in the correct format 'w:h'
+      var isValidFormat = function() {
```
sherwinski: I'm not sure if it's ok stylistically to have anonymous functions littered throughout the code. The main reasons I chose to do it this way was for readability and to avoid exposing certain private functions on the `ImgixClient.prototype`

src/imgix-core-js.js
```diff
@@ -124,6 +124,74 @@
...
+        return this._buildDPRSrcSet(path, params);
+      }
+      else {
+        return this._buildSrcSetPairs(path, params);
+      }
+    };
+
+    ImgixClient.prototype._buildSrcSetPairs = function(path, params) {
+      var srcset = '';
+      var targetWidths = function() {
```
sherwinski: I think that `targetWidths` should definitely be pulled out into a separate file where the results are imported once into this main class. But with the way it's packaged together I'm not exactly sure what's the best way to do that.

src/imgix-core-js.js
jayeb: I'd say you're not gaining anything by using a function here. I'd just do this:

```js
var isValidArFormat = /^\d+(\.\d+)?:\d+(\.\d+)?$/.test(aspectRatio);
```

No unnecessary function scope, no exposed private function, and no litter.

src/imgix-core-js.js
```diff
@@ -124,6 +124,74 @@
...
+      // determines if an aspect ratio value is in the correct format 'w:h'
+      var isValidFormat = function() {
+        return /^\d+(\.\d+)?:\d+(\.\d+)?$/.test(aspectRatio);
+      }();
+
+      if (aspectRatio && !isValidFormat) {
+        throw new Error('The \'ar\' parameter key must follow the format w:h');
+      }
+
+      var fixedWidth = ((width && height) || (width && aspectRatio) || (height && aspectRatio)) ? true : false;
```
jayeb: This also feels contrived to me. Why not just put the condition in the conditional?

```js
// If two of the three are defined, this is a fixed-dimension srcset
if ((width && height) || (width && aspectRatio) || (height && aspectRatio)) {
  return this._buildDPRSrcSet(path, params);
} else {
  return this._buildSrcSetPairs(path, params);
}
```

src/imgix-core-js.js
jayeb: I agree, but I don't think it's worth messing around with the build system at the moment. I'd say for now just pull this out into a "constant" at the top of the file so it doesn't need to be recomputed every time this function is called.

```js
var TARGET_WIDTHS = (function() {
 // ...
})();

```

src/imgix-core-js.js
```diff
@@ -124,6 +124,74 @@
...
+          resolutions.push(ensureEven(prev));
+          prev *= 1 + (INCREMENT_PERCENTAGE / 100) * 2;
+        }
+      
+        resolutions.push(MAX_SIZE);
+        return resolutions;
+      }();
+
+      for(var i = 0; i < targetWidths.length; i++) {
+        currentWidth = targetWidths[i];
```
jayeb: This variables aren't declared in this function, so they're gonna end up on the global scope. We should do a `var currentWidth, currentParams;` at the top of the function (or right before this loop, since we don't seem to care about hoisting variable declarations anywhere else).

sherwinski:
@jayeb Also wanted to ask, do you have any ideas for other test cases that could helpful here?

test/test-client.js
```diff
@@ -335,4 +335,94 @@ describe('Imgix client:', function describeSuite() {
       assert.equal(expectation, result);
     });
   });
+
+  describe('Calling buildSrcSet()', function describeSuite() {
+    describe('width descriptor mode', function describeSuite() {
```
jayeb: Rather than testing this explicitly as "width descriptor mode" and "dpr mode", can we instead test that it behaves correctly under the actual-use scenarios?

- With a `w` parameter provided
- With a `h` parameter provided
- With a `w` and an `h` parameter provided
- ... et cetera

src/imgix-core-js.js
```diff
@@ -124,6 +146,56 @@
       }
     };
 
+    ImgixClient.prototype.buildSrcSet = function (path, params) {
+      var width = params ? params['w'] : undefined;
+      var height = params ? params['h'] : undefined;
+      var aspectRatio = params ? params['ar'] : undefined;
+      if (params) {
+        delete params.dpr;
```
sherwinski: To the best of my knowledge, this would be the correct way to delete a particular key. Just want to be sure that there isn't some sort of 'gotcha' to doing it this way.

jayeb: Nope, this is the way to do it.

PR #59, opened 2019-08-28
fix: include dpr parameter when generating fixed-width srcset

sherwinski:
This PR adds a few things that were missed in #53, namely:
- including the `dpr` parameter when generating a DPR srcset
- mentioning the use of `fit=crop` in documentation
- more test cases for default (no parameter) calls to `buildSrcSet()`, expected srcset pair values, and DPR mode

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -176,10 +176,12 @@
     ImgixClient.prototype._buildDPRSrcSet = function(path, params) {
         var srcset = '';
         var targetRatios = [1, 2, 3, 4, 5];
-        
+        var currentRatio;
+
         for(var i = 0; i < targetRatios.length; i++) {
-          params.dpr = i + 1;
-          srcset += this.buildURL(path, params) + ' ' + targetRatios[i] + 'x,\n'
+          currentRatio = targetRatios[i];
+          params.dpr = currentRatio;
+          srcset += this.buildURL(path, params) + ' ' + currentRatio + 'x,\n'
         }
 
         return srcset.slice(0,-2);
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -147,9 +147,10 @@
     };
 
     ImgixClient.prototype.buildSrcSet = function (path, params) {
-      var width = params ? params['w'] : undefined;
-      var height = params ? params['h'] : undefined;
-      var aspectRatio = params ? params['ar'] : undefined;
+      var params = params || {};
+      var width = params.w;
+      var height = params.h;
+      var aspectRatio = params.ar;
 
       if ((width) || (height && aspectRatio)) {
         return this._buildDPRSrcSet(path, params);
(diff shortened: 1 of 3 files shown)


src/imgix-core-js.js
```diff
@@ -176,11 +176,12 @@
     ImgixClient.prototype._buildDPRSrcSet = function(path, params) {
         var srcset = '';
         var targetRatios = [1, 2, 3, 4, 5];
-        var url = this.buildURL(path, params);
         
         for(var i = 0; i < targetRatios.length; i++) {
           currentRatio = targetRatios[i];
-          srcset += url + ' ' + currentRatio +'x,\n'
+          currentParams = params;
```
jayeb: Couple of problems here:

- `currentParams` isn't declared as a variable, so it's going to be applied to the global scope instead of being scoped to this method.
- It's probably not even worth re-declaring this. We should just use `params`.
- Except we also need to make sure this doesn't break if no `params` are passed in. Something like `params = params || {};`

test/test-client.js
```diff
@@ -338,6 +338,95 @@ describe('Imgix client:', function describeSuite() {
...
+        var max = Number.parseFloat(
+          srcsetSplit[srcsetSplit.length-1]
+          .split(" ")[1]
+          .slice(0, -1)
+        );
+        assert(min >= 100);
+        assert(max <= 8192);
+      });
+
+      it('should not increase more than 18% every iteration', function testSpec() {
```
jayeb: This should be 16%, right?

sherwinski: It should but looks like with the way we're currently rounding, a lot of values come out a little over 16%.

src/imgix-core-js.js
```diff
@@ -176,11 +176,10 @@
     ImgixClient.prototype._buildDPRSrcSet = function(path, params) {
         var srcset = '';
         var targetRatios = [1, 2, 3, 4, 5];
-        var url = this.buildURL(path, params);
         
         for(var i = 0; i < targetRatios.length; i++) {
-          currentRatio = targetRatios[i];
-          srcset += url + ' ' + currentRatio +'x,\n'
+          params.dpr = i + 1;
```
jayeb: Feels weird to use `i + 1` here but `targetRatios[i]` below, when we're actually trying to make the two match up.

test/test-client.js
```diff
@@ -338,6 +338,95 @@ describe('Imgix client:', function describeSuite() {
...
+        var max = Number.parseFloat(
+          srcsetSplit[srcsetSplit.length-1]
+          .split(" ")[1]
+          .slice(0, -1)
+        );
+        assert(min >= 100);
+        assert(max <= 8192);
+      });
+
+      it('should not increase more than 17% every iteration', function testSpec() {
```
jayeb: Will you add a comment here that explains why we're testing for 17%?

PR #64, opened 2019-09-30
feat: add typescript declaration file for ImgixClient

### LINKED ISSUE #50: Typings
Would it be possible to get some typescript typings for this package, either in the repo or on definitely typed?

sherwinski:
**Work in Progress**

This PR adds a typescript declaration file for the `ImgixClient` class. Ideally, these changes should not interfere with users who wish to use this package without typescript.

Resolves #50


sherwinski:
hey @hennessyevan and @gatherben, I've put together a TS declaration file to address #50. There's definitely some other stuff I want to eventually add, such as better typings for `params`, but wanted to get your initial thoughts on this as a starting point. Thanks!

package.json
```diff
@@ -5,14 +5,17 @@
   "repository": "https://github.com/imgix/imgix-core-js",
   "scripts": {
     "assert_version": "node assert_version.js",
-    "compile": "mkdir -p dist && uglifyjs ./src/imgix-core-js.js -o ./dist/imgix-core-js.min.js && cp ./src/imgix-core-js.js ./dist/imgix-core-js.js",
+    "compile": "mkdir -p dist && uglifyjs ./src/imgix-core-js.js -o ./dist/imgix-core-js.min.js && cp ./src/imgix-core-js.js ./dist/imgix-core-js.js && cp ./src/imgix-core-js.d.ts ./dist/imgix-core-js.d.ts && tsc",
     "prepublish": "npm run compile && npm run assert_version",
     "test": "mocha"
   },
   "main": "dist/imgix-core-js.js",
+  "types": "src/",
```
jayeb: Should this be pointing to a specific file, or is it correct to point to the whole directory?

hennessyevan: Should point to `src/imgix-core-js.d.ts` in this case as per https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

sherwinski: > Set the `types` property to point to your bundled declaration file.

Hmm so it would actually have to point to `dist/imgix-core-js.d.ts` in that case, no?

jayeb: Oh yea, good call

sherwinski:
Thanks for the help @hennessyevan @jayeb 👍

PR #108, opened 2020-02-08
feat: add support for defining a min and max srcset width

sherwinski:
This PR adds support to specify a minimum and/or maximum width range when generating `srcset` pairs. In certain situations where the exact min and/or max renderable sizes are known, a user may want to tailor their `srcset` list to be more concise. This is now possible by passing a positive integer value to either the `minWidth` or `maxWidth` parameters, now a part of a new second optional arguent when invoking `buildSrcSet()`. Below is a demonstration of the new argument in action:

```js
  domain:'testing.imgix.net',
  includeLibraryParam: false
  })
var srcset = client.buildSrcSet('image.jpg', {}, {minWidth:500, maxWidth: 2000})
```

This will generate a more specific `srcset` list as follows:

```html
https://testing.imgix.net/image.jpg?w=500 500w,
https://testing.imgix.net/image.jpg?w=580 580w,
https://testing.imgix.net/image.jpg?w=672 672w,
https://testing.imgix.net/image.jpg?w=780 780w,
https://testing.imgix.net/image.jpg?w=906 906w,
https://testing.imgix.net/image.jpg?w=1050 1050w,
https://testing.imgix.net/image.jpg?w=1218 1218w,
https://testing.imgix.net/image.jpg?w=1414 1414w,
https://testing.imgix.net/image.jpg?w=1640 1640w,
https://testing.imgix.net/image.jpg?w=1902 1902w,
https://testing.imgix.net/image.jpg?w=2000 2000w
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@
   - [ES6 Modules](#es6-modules)
   - [In-browser](#in-browser)
 - [Srcset Generation](#srcset-generation)
-- [What is the ixlib param on every request?](#what-is-the-ixlib-param-on-every-request)
+  - [Fixed image rendering](#fixed-image-rendering)
+  - [Minimum and Maximum Width Ranges](#minimum-and-maximum-width-ranges)
+- [What is the `ixlib` param on every request?](#what-is-the-ixlib-param-on-every-request)
 - [Testing](#testing)
 
 ## Installing
@@ -37,7 +39,7 @@ bower install --save imgix-core-js
 
 ## Usage
 
-Depending on your module system, using imgix-core-js is done a few different ways. The most common entry point will be the `Client` class. Whenever you provide data to imgix-core-js, make sure it is not already URL-encoded, as the library handles proper encoding internally.
+Depending on your module system, using imgix-core-js is done a few different ways. The most common entry point will be the `ImgixClient` class. Whenever you provide data to imgix-core-js, make sure it is not already URL-encoded, as the library handles proper encoding internally.
 
 ### CommonJS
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -126,6 +130,40 @@ https://my-social-network.imgix.net/image.jpg?h=800&ar=3%3A2&fit=crop&dpr=5&s=7c
 
 For more information to better understand `srcset`, we highly recommend [Eric Portis' "Srcset and sizes" article](https://ericportis.com/posts/2014/srcset-sizes/) which goes into depth about the subject.
 
+### Minimum and Maximum Width Ranges
+
+If the exact number of minimum/maximum physical pixels that an image will need to be rendered at is known, a user can specify them by passing a positive integer via `minWidth` and/or `maxWidth` to the srcset-modifying parameter:
```
jayeb:
```suggestion
In certain circumstances, you may want to limit the minimum or maximum value of the non-fixed `srcset` generated by the `buildSrcSet()` method. To do this, you can pass in an options object as a third argument, providing positive integers as `minWidth` and/or `maxWidth` attributes:
```

src/imgix-core-js.js
```diff
@@ -187,6 +208,12 @@
         return srcset.slice(0,-2);
     };
 
+    validateRange = function(minSrcset, maxSrcset) {
```
jayeb:
```suggestion
    function validateRange(minSrcset, maxSrcset) {
```

jayeb: Might be clearer to just rename these args to `min` and `max`, as well.

src/imgix-core-js.js
```diff
@@ -17,25 +17,34 @@
...
   // returns an array of width values used during scrset generation
-  var TARGET_WIDTHS = (function() {
+  function _generateTargetWidths(minWidth, maxWidth) {
     var resolutions = [];
-    var prev = 100;
     var INCREMENT_PERCENTAGE = 8;
-    var MAX_SIZE = 8192;
-  
+    var minSize = minWidth;
+    var maxSize = maxWidth;
```
jayeb: Should these values be coerced into integers with a `Math.floor` at this point? It's a little unexpected but right now it seems like there's nothing stopping me from passing `2000.5` as my max width and really screwing things up.

sherwinski: Ah yeah, good catch. I should go back and do the same everywhere else.

PR #109, opened 2020-02-13
feat: add support for defining a custom srcset width tolerance

sherwinski:
This PR adds logic which will allow users to modify the width tolerance used when building `srcset` width pairs. More specifically, the width tolerance dictates the maximum difference between entries in a `srcset` attribute that should be generated. By default, this rate is set to 8 (`0.08`) percent. Users can use this setting to fine tune how many `srcset` pairs they generate when calling `buildSrcSet()`.

A user can now specify their own tolerance rate by passing a positive scalar value to the `widthTolerance` within the method's optional third argument. See the following example:

```js
  domain:'testing.imgix.net',
  includeLibraryParam: false
  })
var srcset = client.buildSrcSet('image.jpg', {}, {widthTolerance: .20})

console.log(srcset);
```
In this case, the `width_tolerance` is set to 20 percent, which will be reflected in the difference between subsequent widths in a `srcset` pair:
```
https://testing.imgix.net/image.jpg?w=100 100w,
https://testing.imgix.net/image.jpg?w=140 140w,
https://testing.imgix.net/image.jpg?w=196 196w,
							...
https://testing.imgix.net/image.jpg?w=8192 8192w
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@
   - [In-browser](#in-browser)
 - [Srcset Generation](#srcset-generation)
   - [Fixed image rendering](#fixed-image-rendering)
+  - [Width Tolerance](#width-tolerance)
   - [Minimum and Maximum Width Ranges](#minimum-and-maximum-width-ranges)
 - [What is the `ixlib` param on every request?](#what-is-the-ixlib-param-on-every-request)
 - [Testing](#testing)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -130,6 +131,32 @@ https://testing.imgix.net/image.jpg?h=800&ar=3%3A2&fit=crop&dpr=5&s=7c4b8adb733d
 
 For more information to better understand `srcset`, we highly recommend [Eric Portis' "Srcset and sizes" article](https://ericportis.com/posts/2014/srcset-sizes/) which goes into depth about the subject.
 
+### Width Tolerance
+
+The `srcset` width tolerance dictates the maximum tolerated size difference between an image's downloaded size and its rendered size. For example: setting this value to 0.1 means that an image will not render more than 10% larger or smaller than its native size. In practice, the image URLs generated for a width-based srcset attribute will grow by twice this rate. A lower tolerance means images will render closer to their native size (thereby reducing rendering artifacts), but a large srcset list will be generated and consequently users may experience lower rates of cache-hit for pre-rendered images on your site.
```
jayeb:
```suggestion
The `srcset` width tolerance dictates the maximum tolerated size difference between an image's downloaded size and its rendered size. For example: setting this value to 0.1 means that an image will not render more than 10% larger or smaller than its native size. In practice, the image URLs generated for a width-based srcset attribute will grow by twice this rate. A lower tolerance means images will render closer to their native size (thereby increasing perceived image quality), but a large srcset list will be generated and consequently users may experience lower rates of cache-hit for pre-rendered images on your site.
```

README.md
```diff
@@ -130,6 +131,32 @@ https://testing.imgix.net/image.jpg?h=800&ar=3%3A2&fit=crop&dpr=5&s=7c4b8adb733d
...
+
+By default this rate is set to 8 percent, which we consider to be the ideal rate for maximizing cache hits without sacrificing visual quality. Users can specify their own width tolerance by providing a positive scalar value as `widthTolerance` to the third options object:
+
+```js
+var client = new ImgixClient({
+  domain:'testing.imgix.net',
+  includeLibraryParam: false
+  })
+var srcset = client.buildSrcSet('image.jpg', {}, {widthTolerance: .20})
```
jayeb:
```suggestion
var srcset = client.buildSrcSet('image.jpg', {}, {widthTolerance: 0.20})
```

src/imgix-core-js.js
```diff
@@ -214,6 +218,12 @@
       }
     };
 
+    function validateWidthTolerance(widthTolerance) {
+      if (typeof widthTolerance != 'number' || widthTolerance < 0) {
```
jayeb: We don't want someone passing a value greater than 1, do we? I don't think it would break anything, but would we ever want to encourage that?

```suggestion
      if (typeof widthTolerance != 'number' || widthTolerance < 0 || widthTolerance > 1) {
```

sherwinski: No I don't think we would ever encourage that. But I am also hesitant to strictly forbid it. 1) I don't want to make assumptions about anyone's use case so as to say they could never want it and 2) As you said nothing will break, but it will probably not be an optimal solution. I think that's where we have drawn the line with validations in the past.

PR #114, opened 2020-02-19
fix: throw error when certain srcset modifiers are passed zero

sherwinski:
This PR patches a hole in the validation that allowed a user to pass a value of `0` to either the `widthTolerance`, `minWidth`, and/or `maxWidth` fields in `createSrcSet()`. Beforehand, users would be able to pass a zero value to these fields - although it would usually end up being ignored. Now the validations are adjusted to throw an error. In addition to this patch, this PR also refactors/cleans-up a lot of the validation logic used in generating `srcset` pairs, pulling it out into its own separate private function. This should yield significantly more readable code.

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -175,8 +175,7 @@
 
       if ((width) || (height && aspectRatio)) {
         return this._buildDPRSrcSet(path, params, options);
-      }
-      else {
+      } else {
         return this._buildSrcSetPairs(path, params, options);
       }
     };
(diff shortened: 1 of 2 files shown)


src/imgix-core-js.js
```diff
@@ -185,21 +184,16 @@
       var srcset = '';
       var currentWidth;
       var targetWidths;
-      var widthTolerance = options["widthTolerance"] || DEFAULT_SRCSET_WIDTH_TOLERANCE;
-      var minWidth = options["minWidth"] || MIN_SRCSET_WIDTH;
-      var maxWidth = options["maxWidth"] || MAX_SRCSET_WIDTH;
-      var customWidths = options["widths"];
+      var customWidths = options.widths;
+      var srcsetOptions = validateAndDestructureOptions(options);
+      var widthTolerance = srcsetOptions[0], minWidth = srcsetOptions[1], maxWidth = srcsetOptions[2];
```
jayeb: So clunky 🤦‍♂. We should consider getting some Babel in here in the near future so we can use some more modern stuff.

src/imgix-core-js.js
```diff
@@ -237,23 +231,50 @@
...
+        maxWidth = MAX_SRCSET_WIDTH;
+        validateRange(minWidth, maxWidth);
+      } else if (options.maxWidth !== undefined) {
+        minWidth = MIN_SRCSET_WIDTH;
+        maxWidth = options.maxWidth;
+        validateRange(minWidth, maxWidth);
+      } else {
+        minWidth = MIN_SRCSET_WIDTH;
+        maxWidth = MAX_SRCSET_WIDTH;
+      }
```
jayeb: I think you may have overcomplicated this.

``` js
minWidth = options.minWidth === undefined ? MIN_SRCSET_WIDTH : options.minWidth;
maxWidth = options.maxWidth === undefined ? MAX_SRCSET_WIDTH : options.maxWidth;

// Validate the range unless we're using defaults for both
if (minWidth != MIN_SRCSET_WIDTH || maxWidth != MAX_SRCSET_WIDTH) {
  validateRange(minWidth, maxWidth);
}
```

src/imgix-core-js.js
```diff
@@ -237,23 +231,50 @@
...
+
+      return [widthTolerance, minWidth, maxWidth];
+    };
+
     function validateRange(min, max) {
-      if (!(Number.isInteger(min) && Number.isInteger(max)) || (min < 0 || max < 0)) {
+      if (!(Number.isInteger(min) && Number.isInteger(max)) || (min <= 0 || max <= 0)) {
```
jayeb:
```suggestion
      if (!(Number.isInteger(min) && Number.isInteger(max)) || (min <= 0 || max <= 0) || min < max) {
```

jayeb: Can we add a test for this new case as well?

sherwinski: I think you mean `min > max` ?

src/imgix-core-js.js
sherwinski: I remember this came up last year but don't recall the exact reasoning for not adding it sooner.

test/test-client.js
```diff
@@ -1074,22 +1074,46 @@ describe('Imgix client:', function describeSuite() {
...
+        });
+
+        it('errors with two invalid widths', function testSpec() {
+          assert.throws(function() {
+            new ImgixClient({
+              domain: 'testing.imgix.net'
+            }).buildSrcSet('image.jpg', {}, {minWidth: 0, maxWidth: 0});
+          }, Error);
+        });
+
```
jayeb: Can we get `it('errors when the maxWidth is less than the minWidth', function testSpec() { ... });`?

PR #115, opened 2020-02-20
perf(srcset): memoize generated srcset width-pairs

sherwinski:
This PR adds memoization to the `_generateTargetWidths` helper function, which is responsible for calculating the widths returned in a [fluid width](https://github.com/imgix/imgix-core-js#srcset-generation) `srcset`.
With memoization, the `ImgixClient` class will build an internal cache of widths as they are generated for the first time, which can then be accessed later on rather than be repeatedly recalculated. Part of this change requires exposing `_generateTargetWidths` on the class prototype, rather than maintaining it privately.


sherwinski:
I am currently on the second page of google results looking for an answer.

PR #127, opened 2020-04-14
test: explicitly test overriding libraryParam for imgix SDKs

frederickfogerty:
I made this PR because we are using this pseudo-public API in our Angular and Vue libraries (and probably others) to set `ixlib` to the correct value for that library. I thought it would be good idea to explicitly state this as a "imgix-public" API that we can depend on in SDK libraries, so we don't end up changing this library and breaking this API, and only realising down the track when we pull this library into other SDKs.


frederickfogerty:
All good! Thanks for the review 🙌

PR #130, opened 2020-04-22
build: drop support for Node.js 8.x

sherwinski:
This PR moves this project away from building for Node.js v8.x as it reached [EOL](https://endoflife.software/programming-languages/server-side-scripting/nodejs) in December 2019. This also offers the secondary benefit of allowing us to utilize more [ECMA2015 features](https://node.green/) moving forward.

Unblocks #99

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 before_install:
 - npm install -g npm@latest
 node_js:
-- '8.0'
 - '10.0'
 - '12.0'
+- '13.0'
+


tremby:
I've only found my way here because I submitted some patches today and felt awkward writing `var`! I don't exactly mean to poke my nose into your tech choices, but it occurs to me that this patch changes the version to node 13. 13 is already end-of-life too -- its support ended in June of this year! The odd-numbered releases are shorter-lived; I imagine it might be better to choose node 12 or 14.

sherwinski:
Good call @tremby. To give additional context here, we are working on giving imgix-core-js a significant update, so this won't be an issue for that much longer. After the point where that is published is when we plan to integrate it into our other libraries like [react-imgix](https://github.com/imgix/react-imgix/issues/723).

sherwinski:
I'm going actually going to close this PR since it will have to be handled as part of the rewrite.

PR #141, opened 2020-06-12
docs: add pull request template

sherwinski:


.github/pull_request_template.md
```diff
@@ -0,0 +1,75 @@
...
+
+<!-- 
+Please use the checklist that is most closely related to your PR, and delete the other checklists. -->
+## Checklist: Fixing typos/Doc change
+
+- [ ] Each commit follows the conventional commit spec format: e.g. `chore(readme): fixed typo`. See the end of this file for more information.
+
+## Checklist: Bug Fix
+
+- [ ] Each commit follows the conventional commit spec format: e.g. `chore(readme): fixed typo`. See the end of this file for more information.
```
ericdeansanchez, line 19: nit: "conventional commit spec format" is a bit of a mouthful and "conventions" "spec," and "format" all mean roughly the same thing in this context. Really, we use this format because it's intended to be _"semantic."_

- "Each commit uses the **semantic format**, e.g. `chore(readme): fix typo`..."

.github/pull_request_template.md
```diff
@@ -0,0 +1,75 @@
...
+  - `refactor`: a code change that neither fixes a bug nor adds a feature
+  - `chore`: reoccurring tasks for project maintainability (example scopes: release, deps)
+  - `config`: changes to tooling configurations used in the project
+  - `build`: changes that affect the build system or external dependencies (example scopes: npm, bundler, gradle)
+  - `ci`: changes to CI configuration files and scripts (example scopes: travis)
+  - `perf`: a code change that improves performance
+  - `style`: changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
+
+`scope` is optional, and can be anything.
+`description` should be a short description of the change, in past tense.
```
ericdeansanchez, line 73: nit: what about the imperative?

Is a change in the past if hasn't been applied?

.github/pull_request_template.md
sherwinski: Just fyi, ["Conventional Commit"](https://www.conventionalcommits.org/en/v1.0.0/#summary) refers to the name of the spec -- but I see how that can be confusing. Maybe it would be better to include a link for users to learn more, if desired.

To your point, I think "[Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format" would be a little more clear.

.github/pull_request_template.md
sherwinski: Good catch 👍

.github/pull_request_template.md
ericdeansanchez: Ah! TIL! Thanks for the link to "conventional commit" I was unaware of this (I assumed there was only semantic-release involvement here, but it's that + conventional commit, cool!)

PR #167, opened 2020-09-30
fix(buildSrcSet): ensure operation is idempotent

### LINKED ISSUE #158: `buildSrcSet` mutates `params` argument when building either fluid or fixed-width srcset.

**Describe the bug**
In the logic used to build the srcset url, the source code assigns extra keys to the "imgix params" object passed to it by the caller.

**To Reproduce**
An example to illustrate the problem, although other possible combinations will exist.

```js
const Imgix = require("imgix-core-js")

const client = new Imgix({
    domain: 'test.imgix.net'
});

const params = {}
const srcsetOptions = { widths: [100] }
const url1 = client.buildSrcSet('', params, srcsetOptions);
const url2 = client.buildSrcSet('', params, srcsetOptions);
console.log(params, url1, url2)
```

image

It can clearly be seen in the screenshot above that params has been mutated. This causes the two urls generated to be different even though they were called with the same parameters.

Repro here: https://runkit.com/frederickfogerty/5f3e2c640856920019d6bfd8

**Expected behavior**
`params` and other parameters to not be mutated

**Screenshots**

**Information:**
 - imgix-core-js version: 2.3.1

**Additional context**
I think the entirety of imgix-core-js should be audited for this problem. I have found the following errors in the source code, but there might be more.

https://github.com/imgix/imgix-core-js/blob/main/src/imgix-core-js.js#L111
https://github.com/imgix/imgix-core-js/blob/main/src/imgix-core-js.js#L179
https://github.com/imgix/imgix-core-js/blob/main/src/imgix-core-js.js#L199
https://github.com/imgix/imgix-core-js/blob/main/src/imgix-core-js.js#L202

ericdeansanchez:
The purpose of this PR is to ensure the `buildSrcSet` operation is
idempotent. Prior to this PR, code modified the input-parameters,
`params`, causing identical calls to `buildSrcSet` to produce different
results.

Now, the input `params` are copied into a new `queryParams` object and
this object is passed to callers requiring `params`. A test has been
written to show that the behavior has changed from the behavior detailed
in issue #158––the input `params` object remains unchanged after calling
`buildSrcSet` _and_ that calling `buildSrcSet` multiple times produces
the same result (given the same inputs).

I also ripgrep'd through the repo with `rg 'params.[:alpha:]'` and with
`rg 'params\['` to ensure `params` is never mutated (i.e. it never
appears on the left hand side of any expression).

Closes #158

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -107,11 +107,11 @@
     };
 
     ImgixClient.prototype._buildParams = function(params) {
+      var queryParams = [];
       if (this.settings.libraryParam) {
-        params.ixlib = this.settings.libraryParam
+        queryParams.push('ixlib=' + this.settings.libraryParam)
       }
 
-      var queryParams = [];
       var key, val, encodedKey, encodedVal;
       for (key in params) {
         val = params[key];
@@ -174,10 +174,16 @@
         targetWidths = this._generateTargetWidths(widthTolerance, minWidth, maxWidth);
       }
 
+      var queryParams = {};
+      var key;
+      for (key in params) {
+        queryParams[key] = params[key];
+      }
+
       for (var i = 0; i < targetWidths.length; i++) {
         currentWidth = targetWidths[i];
-        params.w = currentWidth;
-        srcset += this.buildURL(path, params) + ' ' + currentWidth + 'w,\n';
+        queryParams.w = currentWidth;
+        srcset += this.buildURL(path, queryParams) + ' ' + currentWidth + 'w,\n';
       }
 
       return srcset.slice(0,-2);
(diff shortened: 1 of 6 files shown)


src/imgix-core-js.js
```diff
@@ -174,10 +174,16 @@
         targetWidths = this._generateTargetWidths(widthTolerance, minWidth, maxWidth);
       }
 
+      var queryParams = {};
+      var key, value;
```
frederickfogerty, line 178: I think value is unused here

test/test-buildSrcSet.js
```diff
@@ -387,6 +387,21 @@ describe('SrcSet Builder:', function describeSuite() {
                         assert(src.includes(`q=${QUALITY_OVERRIDE}`));
                     }
                 });
+
+                it('should not modify input params and should be idempotent', function testSpec() {
```
frederickfogerty, line 391: Could we add the same test for `.buildSrc` too?

frederickfogerty (changes requested):
Looks good @ericdeansanchez! Only a few small things, overall looks great ☺️

frederickfogerty:
No change needed and it's not a big thing, but we don't usually include the changes to `dist` in a PR, and usually only commit them when doing a release.

frederickfogerty:
Also it looks like you're going to need to rebase onto main before merging to keep the history clean 😄

src/imgix-core-js.js
ericdeansanchez: Yeah, you're right, good catch! I wrote this in two places (L197, L177), but forgot to remove the `value` var here. Let me clean this up.

test/test-buildSrcSet.js
ericdeansanchez: Yessir!

ericdeansanchez:
Thanks for the great feedback! I think what happened with `dist` was I ran `npm install` after making a couple changes. Then I wondered if changing `dist` was preferred/or not, but you've answered that question. TIL

src/imgix-core-js.js
```diff
@@ -174,10 +174,16 @@
...
+
       for (var i = 0; i < targetWidths.length; i++) {
         currentWidth = targetWidths[i];
-        params.w = currentWidth;
-        srcset += this.buildURL(path, params) + ' ' + currentWidth + 'w,\n';
+        queryParams.w = currentWidth;
+        srcset += this.buildURL(path, queryParams) + ' ' + currentWidth + 'w,\n';
       }
 
       return srcset.slice(0,-2);
```
sherwinski, line 189: Nit: I know this outside of the scope of this change but what if we initialized the local var `srcset` as an array and then `.join("\n")` at the end?

ericdeansanchez: Yeah, I plan to do that, but didn't because it was outside the scope of this change/issue. 

This is something I've had my eye on for a while, but planned to follow this PR with another to make these changes.

However, I am more than happy to make them within this PR though. Just lmk :)

ericdeansanchez:
I'm closing this PR in favor of `fix(buildURL): ensure operation is idempotent` #168.  

In the above PR, I tried to use git to clean up some commits, but there were a couple casualty-commits (that were undone unfortunately :upside_down_face:.

PR #168, opened 2020-10-01
fix(buildURL): ensure operation is idempotent

### LINKED ISSUE #158: `buildSrcSet` mutates `params` argument when building either fluid or fixed-width srcset.

ericdeansanchez:
The purpose of this PR is to ensure the `buildSrcSet` operation is
idempotent. Prior to this PR, code modified the input-parameters,
`params`, causing identical calls to `buildSrcSet` to produce different
results.

Now, the input `params` are copied into a new `queryParams` object and
this object is passed to callers requiring `params`. A test has been
written to show that the behavior has changed from the behavior detailed
in issue #158––the input `params` object remains unchanged after calling
`buildSrcSet` _and_ that calling `buildSrcSet` multiple times produces
the same result (given the same inputs).

I also ripgrep'd through the repo with `rg 'params.[:alpha:]'` and with
`rg 'params\['` to ensure `params` is never mutated (i.e. it never
appears on the left hand side of any expression).

Tests have been added to ensure that `_buildParams`:
- does not modify its input-argument

Tests have been added to ensure that calling `buildURL`:
- is an idempotent operation with empty args
- is an idempotent operation with args
- does not modify empty args
- does not modify its args (non-empty)

Tests have been added to ensure that `buildSrcSet`:
- does not modify input params and is idempotent

Closes #158

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -107,11 +107,11 @@
     };
 
     ImgixClient.prototype._buildParams = function(params) {
+      var queryParams = [];
       if (this.settings.libraryParam) {
-        params.ixlib = this.settings.libraryParam
+        queryParams.push('ixlib=' + this.settings.libraryParam)
       }
 
-      var queryParams = [];
       var key, val, encodedKey, encodedVal;
       for (key in params) {
         val = params[key];
@@ -174,10 +174,16 @@
         targetWidths = this._generateTargetWidths(widthTolerance, minWidth, maxWidth);
       }
 
+      var key;
+      var queryParams = {};
+      for (key in params) {
+        queryParams[key] = params[key];
+      }
+
       for (var i = 0; i < targetWidths.length; i++) {
         currentWidth = targetWidths[i];
-        params.w = currentWidth;
-        srcset += this.buildURL(path, params) + ' ' + currentWidth + 'w,\n';
+        queryParams.w = currentWidth;
+        srcset += this.buildURL(path, queryParams) + ' ' + currentWidth + 'w,\n';
       }
 
       return srcset.slice(0,-2);
@@ -186,23 +192,29 @@
     ImgixClient.prototype._buildDPRSrcSet = function(path, params, options) {
         var srcset = '';
         var targetRatios = [1, 2, 3, 4, 5];
-        var currentRatio;
         var disableVariableQuality = options.disableVariableQuality || false;
-        var quality = params.q;
+
+        var key;
+        var queryParams = {};
+        for (key in params) {
+          queryParams[key] = params[key];
+        }
+
+        var quality = queryParams.q;
 
         if (!disableVariableQuality) {
           validateVariableQuality(disableVariableQuality);
         }
 
         for (var i = 0; i < targetRatios.length; i++) {
-          currentRatio = targetRatios[i];
-          params.dpr = currentRatio;
+          var currentRatio = targetRatios[i];
+          queryParams.dpr = currentRatio;
 
           if (!disableVariableQuality) {
-            params.q = quality || DPR_QUALITIES[currentRatio];
+            queryParams.q = quality || DPR_QUALITIES[currentRatio];
           }
 
-          srcset += this.buildURL(path, params) + ' ' + currentRatio + 'x,\n'
+          srcset += this.buildURL(path, queryParams) + ' ' + currentRatio + 'x,\n'
         }
 
         return srcset.slice(0,-2);
--- a/test/test-buildSrcSet.js
+++ b/test/test-buildSrcSet.js
@@ -387,6 +387,21 @@ describe('SrcSet Builder:', function describeSuite() {
                         assert(src.includes(`q=${QUALITY_OVERRIDE}`));
                     }
                 });
+
+                it('should not modify input params and should be idempotent', function testSpec() {
+                    var client = new ImgixClient({ domain: 'test.imgix.net' });
+                    var params = {};
+                    var srcsetOptions = { widths: [100] };
+                    var srcset1 = client.buildSrcSet('', params, srcsetOptions);
+                    var srcset2 = client.buildSrcSet('', params, srcsetOptions);
+
+                    // Show idempotent, ie. calling buildSrcSet produces the same result given
+                    // the same input-parameters.
+                    assert(srcset1 == srcset2);
+
+                    // Assert that the object remains unchanged.
+                    assert(Object.keys(params).length === 0 && params.constructor === Object);
+                });
             });
 
             describe('with an aspect ratio parameter provided', function describeSuite() {
--- a/test/test-buildURL.js
+++ b/test/test-buildURL.js
@@ -223,11 +223,31 @@ describe('URL Builder:', function describeSuite() {
             assert.equal(expectation, result);
         });
 
+        it('does not modify its input-argument', function testSpec() {
+            var params = {
+                    w: 400,
+                    h: 300
+                },
+                expectation = '?w=400&h=300',
+                result = client._buildParams(params);
+
+            assert.equal(params.w, 400);
+            assert.equal(params.h, 300);
+            assert.equal(expectation, result);
+
+            var emptyParams = {};
+            var emptyResult = client._buildParams(emptyParams);
+
+            // Ensure the result is empty.
+            assert.equal(emptyResult, '');
+            assert(Object.keys(emptyParams).length === 0 && emptyParams.constructor === Object);
+        });
+
         it('includes an `ixlib` param if the `libraryParam` setting is truthy', function testSpec() {
             var params = {
                     w: 400
                 },
-                expectation = '?w=400&ixlib=test',
+                expectation = '?ixlib=test&w=400',
                 result;
 
             client.settings.libraryParam = 'test';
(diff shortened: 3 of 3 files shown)


sherwinski (approved):

test/test-buildSrcSet.js
```diff
@@ -387,6 +387,21 @@ describe('SrcSet Builder:', function describeSuite() {
...
+                    var srcsetOptions = { widths: [100] };
+                    var srcset1 = client.buildSrcSet('', params, srcsetOptions);
+                    var srcset2 = client.buildSrcSet('', params, srcsetOptions);
+
+                    // Show idempotent, ie. calling buildSrcSet produces the same result given
+                    // the same input-parameters.
+                    assert(srcset1 == srcset2);
+
+                    // Assert that the object remains unchanged.
+                    assert(Object.keys(params).length === 0 && params.constructor === Object);
```
frederickfogerty, line 403: Nit: the `params.constructor` seems overkill 😄

test/test-buildURL.js
```diff
@@ -223,11 +223,31 @@ describe('URL Builder:', function describeSuite() {
...
+
+            assert.equal(params.w, 400);
+            assert.equal(params.h, 300);
+            assert.equal(expectation, result);
+
+            var emptyParams = {};
+            var emptyResult = client._buildParams(emptyParams);
+
+            // Ensure the result is empty.
+            assert.equal(emptyResult, '');
```
frederickfogerty, line 242: "Ensure the result is empty" does not seem like a valid assertion given the name of the test - I suggest either to update the test name or drop the assertion (my opinion would be drop the assertion).

test/test-buildURL.js
```diff
@@ -223,11 +223,31 @@ describe('URL Builder:', function describeSuite() {
...
+            var params = {
+                    w: 400,
+                    h: 300
+                },
+                expectation = '?w=400&h=300',
+                result = client._buildParams(params);
+
+            assert.equal(params.w, 400);
+            assert.equal(params.h, 300);
+            assert.equal(expectation, result);
```
frederickfogerty, line 236: Likewise, these assertions don't seem to match the test case. I would recommend dropping them.

test/test-buildURL.js
```diff
@@ -223,11 +223,31 @@ describe('URL Builder:', function describeSuite() {
...
+
+            var emptyParams = {};
+            var emptyResult = client._buildParams(emptyParams);
+
+            // Ensure the result is empty.
+            assert.equal(emptyResult, '');
+            assert(Object.keys(emptyParams).length === 0 && emptyParams.constructor === Object);
+        });
+
         it('includes an `ixlib` param if the `libraryParam` setting is truthy', function testSpec() {
```
frederickfogerty, line 246: I know it's not part of the PR, but if you feel like it, I think this test's assertion could be better as 

```js
assert(result.match(/ixlib=test/));
```

test/test-buildURL.js
```diff
@@ -294,4 +314,61 @@ describe('URL Builder:', function describeSuite() {
...
+    describe('Calling buildURL()', function describeSuite() {
+        var client,
+            path = 'images/1.png';
+
+        beforeEach(function setupClient() {
+            client = new ImgixClient({
+            domain: 'test.imgix.net',
+            includeLibraryParam: false
+            });
+        });
```
frederickfogerty, line 327: No change needed, but I prefer to use `beforeEach` (and other lifecycle methods) strictly for things like setting up DB connections and things. The key thing is that imo, they are best used for things that aren't "coupled" with the test. The client in this case, however, is coupled with the test, since for example the domain is being checked in the "idempotent operation" test. Therefore, I like to instead write these "setup" calls as just functions, and call them in the tests. 

For example, if there was a function like such:

```js
const setupClient = (config) => new ImgixClient({ domain: 'test.imgix.net', includeLibraryParam: false, ...config });
```

then the first test could be:

```js
it('is an idempotent operation with empty args', function testSpec() {
  const client = setupClient();
  const result1 = client.buildURL('', {});
  const result2 = client.buildURL('', {});
  assert.equal(result1, result2);
});
```

The main benefit for me is that the test is more isolated to me, and when reading the test, everything that is relevant to that test (apart from DB connection stuff, etc) is written in that test, and I don't have to go back and forth between different places in the code 😄 

An additional benefit this has is that things like the config can be modified on a test-by-test basis. Otherwise in some tests the client invariably ends up being created _again_, resulting in more code duplication, rather than less 😅

src/imgix-core-js.js
```diff
@@ -186,23 +192,29 @@
...
         var targetRatios = [1, 2, 3, 4, 5];
-        var currentRatio;
         var disableVariableQuality = options.disableVariableQuality || false;
-        var quality = params.q;
+
+        var key;
+        var queryParams = {};
+        for (key in params) {
+          queryParams[key] = params[key];
+        }
```
frederickfogerty, line 201: Nit because I can. So in ES2015 this kind of behaviour is provided by the function "Object.assign". We can't use it (yet) because it's not supported in IE 11, but maybe this could be written as

```suggestion
var queryParams = assign({}, params);
```

and then have a helper method somewhere else that is:

```js
var assign = function assign() {
  var target = arguments[0];
  var rest = arguments.slice(1);

  rest.map(function (obj) { 
    for (key in obj) {
      target[key] = obj[key];
    }
  }); 

  return target
} 
```

Then the code in the for loop can be better represented as (since e.g. the assigning to `queryParams.dps` is a bit of a smell):

```js
for (var i = 0; i < targetRatios.length; i++) {
  var currentRatio = targetRatios[i];

  if (!disableVariableQuality) {
    var q = quality || DPR_QUALITIES[currentRatio];
  }

  srcset +=
    this.buildURL(path, assign({}, queryParams, { dpr: currentRatio, q: q })) + " " + currentRatio + "x,\n";
}
```

frederickfogerty (approved):
Looks good overall! Good work @ericdeansanchez!

test/test-buildURL.js
ericdeansanchez: Cool! Yeah, tbh I wasn't quite sure about the purpose of `beforeEach`, but tried to extrapolate a bit from the usage in this repo. This is good to know, I'll continue to think about this one.

test/test-buildURL.js
ericdeansanchez: I though the goal of this test was to ensure it does not modify its input? I mean, it's pretty trivial, but perhaps redundant?

test/test-buildSrcSet.js
ericdeansanchez: Yeah, I think so, but wasn't sure haha

src/imgix-core-js.js
ericdeansanchez: Yeah, I would have reached for `Object.assign` if it was an option. I would also agree that (most) assignment is a bit of a smell. Although there's a cleaner way of doing the above, I figured we could save that for a future refactor?

It also feels like `sequence.map(forLoop)` may also be a bit of a smell, but then again I'm not sure if this is something that's done frequently in JS. Lmk! ha

test/test-buildURL.js
ericdeansanchez: Agreed. I think I left this in because it was interesting. I wasn't sure if this was documented behavior? I think this should go into a new test (or perhaps just dropped).

src/imgix-core-js.js
frederickfogerty: Yeah you're right `sequence.map(forLoop)` is a bit of a smell, maybe you could use a `.reduce` instead. 

```js
var assign = function assign() {
  return (arguments || []).reduce(function (p, v, i) {
    if (i === 0) { return v };
	Object.keys(v).map(function (key) { p[key] = v[key });
    return p;
  }, {});
} 
```

We could also use the official polyfill from MDN:
```js
if (typeof Object.assign !== 'function') {
  // Must be writable: true, enumerable: false, configurable: true
  Object.defineProperty(Object, "assign", {
    value: function assign(target, varArgs) { // .length of function is 2
      'use strict';
      if (target === null || target === undefined) {
        throw new TypeError('Cannot convert undefined or null to object');
      }

      var to = Object(target);

      for (var index = 1; index < arguments.length; index++) {
        var nextSource = arguments[index];

        if (nextSource !== null && nextSource !== undefined) { 
          for (var nextKey in nextSource) {
            // Avoid bugs when hasOwnProperty is shadowed
            if (Object.prototype.hasOwnProperty.call(nextSource, nextKey)) {
              to[nextKey] = nextSource[nextKey];
            }
          }
        }
      }
      return to;
    },
    writable: true,
    configurable: true
  });
}
```

test/test-buildURL.js
ericdeansanchez: I cleaned this up.

src/imgix-core-js.js
ericdeansanchez: Small follow up note. I think, with the way the function interfaces currently work together , assigning to `queryParams.dps` is necessary because `this.buildURL(path, queryParams)` (ultimately `.__buildParams`) uses the key-value pairs in the `queryParams` object to build the query parameter string.

I think there's a more functional approach that eliminates assignment in a few places, but I figured we'd leave that for a future refactor.

PR #171, opened 2020-10-06
refactor: push-and-join instead of concat-and-slice

ericdeansanchez:
The purpose of this PR is to make `._buildSrcSetPairs` and
`._buildDPRSrcSet` more direct. Before, this PR both functions
were concatenating strings within a loop. In both functions, the
last URL string to be generated includes `",\n"` which is sliced off
when the functions return.

Now, both functions generate/build URLs that are pushed onto the
`srcset` array. The URLs in the `srcset` array are then joined together
with `",\n"`.

This change was made in an attempt to improve the semantics. E.g.
"build URLs and join them with ',\n'" versus "build URLs, concatenate
them, and remove the extra ',\n' from the result".

### DIFF
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -158,7 +158,7 @@
     };
 
     ImgixClient.prototype._buildSrcSetPairs = function(path, params, options) {
-      var srcset = '';
+      var srcset = [];
       var currentWidth;
       var targetWidths;
       var customWidths = options.widths;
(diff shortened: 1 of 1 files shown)


src/imgix-core-js.js
```diff
@@ -158,7 +158,7 @@
     };
 
     ImgixClient.prototype._buildSrcSetPairs = function(path, params, options) {
-      var srcset = '';
```
frederickfogerty, line 161: No change needed, but I would probably call this `srcsets` rather than just `srcset`

frederickfogerty (commented):
This is a good change @ericdeansanchez 👏

frederickfogerty (approved):

ericdeansanchez:
Thanks!

PR #172, opened 2020-10-09
chore: correct typescript-related dependencies

frederickfogerty:
I removed the peerDep on `@types/node` since it's not actually required, and updated the typescript peer dep to work with the oldest version of the type definitions that I could verify worked.

While creating this PR, I realised that we don't actually need a peer dependency on typescript since it a) forces non-TS consumers to install typescript as a dependency (otherwise they will have a persistent warning), and b) since our types work with TS versions all the way back to 1.5, we don't need to restrict the TS version that's installed, which makes the peer dependency redundant. Therefore I removed it.


sherwinski (approved):
Nice! Thanks for doing this @frederickfogerty 👍 👍 

Also, I'm going to go ahead and merge this now so that we can include it in the new release today.

PR #185, opened 2020-11-04
Improve spacing between srcset widths

tremby:
# Description

In this changeset I derive an improved increment ratio for srcset generation from the width tolerance.

Before this patch the widths increase at a regularly-exponential rate until the second-largest width, then the largest width is equal to the maximum width. In all but some very particular cases this will mean the largest two widths are much more closely spaced than the others, in terms of the exponential scale at least. It is possible to have candidate widths of 8190 and 8192 pixels, for example, if the right minWidth, maxWidth, and widthTolerance values are chosen. (Obviously this is an extreme case.)

## Checklist: Bug Fix

- [x] Each commit follows the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: e.g. `chore(readme): fixed typo`. See the end of this file for more information.
- [x] All existing unit tests are still passing (if applicable)
- [x] Add new passing unit tests to cover the code introduced by your PR **(no brand-new tests, but adjusted affected existing ones and added a new clause to the test I added as part of #184 since the new code is no longer susceptible to the previously-failing case)**
- [x] Update the readme
- [ ] Update or add any necessary API documentation **(n/a)**
- [x] Add some [steps](#steps-to-test) so we can test your bug fix

## Steps to Test

I think the changes to the tests already cover this, but as a realistic example where the difference is easier to see:

    client.buildSrcSet('image.jpg', {}, {widthTolerance: 0.1, minWidth: 320, maxWidth: 1440})

- Before the change:
  320,384,460,552,664,796,956,1146,1376,1440

      1600 +----------------------------------------------------------------------+
           |             +             +              +             +             |
           |                                                                      |
      1400 |-+                                                             A    +-|
           |                                                        A             |
           |                                                                      |
           |                                                                      |
      1200 |-+                                               A                  +-|
           |                                                                      |
           |                                                                      |
      1000 |-+                                                                  +-|
           |                                          A                           |
           |                                                                      |
       800 |-+                                 A                                +-|
           |                                                                      |
           |                           A                                          |
       600 |-+                                                                  +-|
           |                    A                                                 |
           |                                                                      |
           |             A                                                        |
       400 |-+    A                                                             +-|
           |                                                                      |
           |             +             +              +             +             |
           +----------------------------------------------------------------------+

  (this plot looks more bumpy than it really is due to ascii; the only important part is the sudden change in direction for the last value)

- After the change:
  320,378,448,528,624,738,872,1030,1218,1440

      1600 +----------------------------------------------------------------------+
           |             +             +              +             +             |
           |                                                                      |
      1400 |-+                                                             A    +-|
           |                                                                      |
           |                                                                      |
           |                                                        A             |
      1200 |-+                                                                  +-|
           |                                                                      |
           |                                                 A                    |
      1000 |-+                                                                  +-|
           |                                                                      |
           |                                          A                           |
       800 |-+                                                                  +-|
           |                                   A                                  |
           |                                                                      |
       600 |-+                         A                                        +-|
           |                                                                      |
           |                    A                                                 |
           |             A                                                        |
       400 |-+    A                                                             +-|
           |                                                                      |
           |             +             +              +             +             |
           +----------------------------------------------------------------------+

## Note

I've built this on top of #184 since that was a bug I found while writing this and they touch some of the same code. As such, you may not need to review all commits; only the last two are strictly relevant. I can rebase it to be straight on top of main if you like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ https://testing.imgix.net/image.jpg?w=100&s=e2e581a39c917bdee50b2f8689c30893 100
 https://testing.imgix.net/image.jpg?w=116&s=836e0bc15da2ad74af8130d93a0ebda6 116w,
 https://testing.imgix.net/image.jpg?w=134&s=688416d933381acda1f57068709aab79 134w,
                                             ...
-https://testing.imgix.net/image.jpg?w=7400&s=91779d82a0e1ac16db04c522fa4017e5 7400w,
+https://testing.imgix.net/image.jpg?w=7074&s=91779d82a0e1ac16db04c522fa4017e5 7074w,
 https://testing.imgix.net/image.jpg?w=8192&s=59eb881b618fed314fe30cf9e3ec7b00 8192w
 ```
 
@@ -259,11 +259,10 @@ console.log(srcset);
 In this case, the `width_tolerance` is set to 20 percent, which will be reflected in the difference between subsequent widths in a srcset pair:
 
 <!-- prettier-ignore-start -->
-
 ```html
 https://testing.imgix.net/image.jpg?w=100 100w,
-https://testing.imgix.net/image.jpg?w=140 140w,
-https://testing.imgix.net/image.jpg?w=196 196w,
+https://testing.imgix.net/image.jpg?w=136 136w,
+https://testing.imgix.net/image.jpg?w=188 188w,
 							...
 https://testing.imgix.net/image.jpg?w=8192 8192w
 ```
@@ -292,16 +291,16 @@ Will result in a smaller, more tailored srcset.
 
 ```html
 https://testing.imgix.net/image.jpg?w=500 500w,
-https://testing.imgix.net/image.jpg?w=580 580w,
-https://testing.imgix.net/image.jpg?w=672 672w,
-https://testing.imgix.net/image.jpg?w=780 780w,
-https://testing.imgix.net/image.jpg?w=906 906w,
-https://testing.imgix.net/image.jpg?w=1050 1050w,
-https://testing.imgix.net/image.jpg?w=1218 1218w,
-https://testing.imgix.net/image.jpg?w=1414 1414w,
-https://testing.imgix.net/image.jpg?w=1640 1640w,
-https://testing.imgix.net/image.jpg?w=1902 1902w,
-https://testing.imgix.net/image.jpg?w=2000 2000w
+https://testing.imgix.net/image.jpg?w=574 574w,
+https://testing.imgix.net/image.jpg?w=660 660w,
+https://testing.imgix.net/image.jpg?w=758 758w,
+https://testing.imgix.net/image.jpg?w=870 870w,
+https://testing.imgix.net/image.jpg?w=1000 1000w,
+https://testing.imgix.net/image.jpg?w=1148 1148w,
+https://testing.imgix.net/image.jpg?w=1320 1320w,
+https://testing.imgix.net/image.jpg?w=1516 1516w,
+https://testing.imgix.net/image.jpg?w=1742 1742w,
+https://testing.imgix.net/image.jpg?w=2000 2000w,
 ```
 
 Remember that browsers will apply a device pixel ratio as a multiplier when selecting which image to download from a `srcset`. For example, even if you know your image will render no larger than 1000px, specifying `options: { max_srcset: 1000 }` will give your users with DPR higher than 1 no choice but to download and render a low-resolution version of the image. Therefore, it is vital to factor in any potential differences when choosing a minimum or maximum range.
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -243,10 +243,26 @@
         return 2 * Math.round(n / 2);
       };
 
+      // Calculate the (probably non-integer) number of increments we'd make
+      // with the requested width tolerance to get from minWidth to maxWidth,
+      // and round it up to ensure the tolerance will be met or beaten.
+      // This formula is found by solving for numSteps the following identity:
+      //   minWidth * INCREMENT_RATIO ** numSteps == maxWidth
+      var numSteps = Math.ceil(Math.log(maxWidth / minWidth) / Math.log(INCREMENT_RATIO));
+
+      // Calculate an adjusted value for INCREMENT_RATIO
+      // such that numSteps would be an integer without the rounding used above.
+      // This is found by solving the same identity as above,
+      // but this time for INCREMENT_RATIO.
+      var adjustedIncrementRatio = Math.pow(maxWidth / minWidth, 1 / numSteps);
+      // Note that this will be NaN if numSteps was zero,
+      // but this is only when minWidth and maxWidth are equal,
+      // in which case adjustedIncrementRatio will not be used.
+
       var exact = minWidth;
       while (resolutions[resolutions.length - 1] < maxWidth) {
-        exact *= INCREMENT_RATIO;
-        resolutions.push(Math.min(ensureEven(exact), maxWidth));
+        exact *= adjustedIncrementRatio;
+        resolutions.push(Math.round(exact) === maxWidth ? maxWidth : ensureEven(exact));
       }
 
       this.targetWidthsCache[cacheKey] = resolutions;
--- a/src/imgix-core-js.js
+++ b/src/imgix-core-js.js
@@ -225,7 +225,7 @@
     // returns an array of width values used during scrset generation
     ImgixClient.prototype._generateTargetWidths = function(widthTolerance, minWidth, maxWidth) {
       var resolutions = [minWidth];
-      var INCREMENT_PERCENTAGE = widthTolerance;
+      var INCREMENT_RATIO = 1 + (widthTolerance * 2);
       var minWidth = Math.floor(minWidth);
       var maxWidth = Math.floor(maxWidth);
 
@@ -233,7 +233,7 @@
         return resolutions;
       }
 
-      var cacheKey = INCREMENT_PERCENTAGE + '/' + minWidth + '/' + maxWidth;
+      var cacheKey = widthTolerance + '/' + minWidth + '/' + maxWidth;
 
       if (cacheKey in this.targetWidthsCache) {
         return this.targetWidthsCache[cacheKey];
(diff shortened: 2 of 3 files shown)


tremby:
I just identified a bug with this, where if I specify an odd maxWidth I can get a repeated value. Working on it.

tremby:
I take it from the lack of comments so far that this hasn't been reviewed yet, so I'm going to meld my fix to the existing commits.

tremby:
The failing set of cases included min 1000, max 1001. In this case it decided 1 step is necessary, decided on the increment factor, then calculated the next stop as being (in this case) something like 1000.99999 due to floating point error. Then `ensureEven` rounded that to 1000, which was the same as the previous width we stored in the array.

I fixed this with a check for whether the running fractional width rounds to the target width. If so, it doesn't run `ensureEven`.

frederickfogerty:
Hey @tremby! @imgix would really like to thank you so much for the activity on this repo! Unfortunately this is a change that we are not on board with, and I'd like to explain a bit why, because we value our contributor's contributions and don't want you to think that we are just dismissing this without reason. 

Firstly, we place large weight on changes to this behaviour, since this would require changes to ~20 of our libraries, since we require 100% consistency across all our SDK libraries for the srcset generation feature. This is important for cache performance.

Mainly, I think the fundamental problem with this PR for us is that it shifts the goal of the srcset feature away from the goal it was designed for, to an arbitrary goal. Our original goal with this srcset feature was to achieve a **suitable trade-off between visual performance, bandwidth usage, and cache hits.** This is the reasoning between the width tolerance of 8%, which was very intentionally chosen.  Therefore, it's important for us that the widths scale up according to this increment percentage every time, because this will ensure optimal performance across the majority (99%) of browser widths and images. 

If we were to adopt this PR, we would be shifting our goal towards something like "smoothness of the srcset widths curve", or "making sure that every value is correctly spaced from every other value", rather than the aforementioned goal. This is not a trade-off we are willing to make, as we value the original goal of this feature, even if that causes small irregularities.

We do understand that the issue you raised is an issue/"bug" with our srcset-generation logic, but as with most problems, no solution is perfect, and in this case this is a trade-off we feel that should make.

Furthermore, I think the irregularity at the top end of the srcset range is a small issue since it will only affect a very small subset of users. Also, I think the goal of optimizing for this 100-8192px range is the wrong way to look at it. Honestly the 8192px hard limit is a constraint of the imgix API, and this will be changed in the future. It would be bad for cache performance in the future that when we change this, all the widths would change and thus all the caches would be invalidated. Instead, it is more important for us that these images remain in the cache, independent of this hard limit. 

And, finally, another small reason why I didn't like this change is because it makes the program not directly follow the developer's input. Right now, the width tolerance is very direct and easy to understand for developers - "If I set width_tolerance to 0.10, that means each width will increase by 10% every time", whereas with this change, they would no longer have control over how much the widths would change every time, which could be frustrating, and we don't want to take this clarity away from them.

Anyway! That was a bit of a wall of text, but I would like to thank you again for investigating this real issue in our logic, and working on a fix for us! We really do appreciate it, and we are welcome to review/discuss any other changes in the future that you might think would improve our libraries.

tremby:
I'm rather confused about this decision.

> this would require changes to ~20 of our libraries, since we require 100% consistency across all our SDK libraries for the srcset generation feature.

Show me the libraries and I could do the legwork for you!

> This is the reasoning between the width tolerance of 8%, which was very intentionally chosen.

But that's just a default, and it's overrideable by the developer. In my opinion it's a much-too-small default (I'd be interested to understand exactly how 8% was decided upon), but that's not the point anyway. The fact is that **my changes result in the exact same number of steps being produced as before the change**. So I'm not sure what you're getting at here.

> Therefore, it's important for us that the widths scale up according to this increment percentage every time, because this will ensure optimal performance across the majority (99%) of browser widths and images.

I don't follow your logic here at all. The developer chooses the `minWidth`, `maxWidth`, and `widthTolerance`. In light of that, how does setting the default tolerance to 8% "ensure optimal performance"? Why would the same number of steps but smoothly spaced (which would perhaps mean the step is 7.5% rather than 8% for a given `minWidth` and `maxWidth`) change that performance?

(Though I absolutely agree that performance is important (and don't think this change is harming it in any way) I can't help but be surprised about the level of weight you're putting on it here given both the (very low in my opinion) 8% default tolerance and the categorical ignorance of cacheability and performance over in your React background image component! https://github.com/imgix/react-imgix/issues/672)

> If we were to adopt this PR, we would be shifting our goal towards something like "smoothness of the srcset widths curve", or "making sure that every value is correctly spaced from every other value", rather than the aforementioned goal.

Again I don't follow. I don't see how this change affects "visual performance, bandwidth usage, and cache hits" other than if somebody updates and so their srcset changes, and so there are some cache misses the first time each size is requested. Again, I'll point out that the same number of steps are being generated under this algorithm. The way I see it your initial goals are still all met, and additionally the spacings are smooth, which can only be an improvement.

> Furthermore, I think the irregularity at the top end of the srcset range is a small issue since it will only affect a very small subset of users.

Surely nobody intentionally leaves `maxWidth` at the default of 8192 pixels? As I've mentioned in other tickets, the default settings cause most images to be upscaled enormously, which no matter which way you look at it is a horrible waste of resources both on your server side and for the user.

The irregularity at the top end of the srcset range would affect many more users where `maxWidth` has been set to something reasonable (i.e. the actual width of the source image). So I don't think the bug should be discounted for the reason you give.

> Also, I think the goal of optimizing for this 100-8192px range is the wrong way to look at it.

There's no optimization for that particular range in this code. I wonder if you misunderstood how the code works?

1. Determine how many steps would be generated using the given `widthTolerance` precisely -- this is the number of steps the library currently generates before this patch. This is fast, done with a couple of logarithms; no iteration.
2. Adjust the scale factor to produce that same number of steps smoothly. This is also fast, done with an exponent (nth-root).

Nothing to do with the 100-8192 range is hardcoded.

> It would be bad for cache performance in the future that when we change this, all the widths would change and thus all the caches would be invalidated.

This is fair enough, and would affect existing images for users who upgrade their library. But it's not like everybody's going to do so, and certainly not simultaneously. It's also a one-time thing. You know your infrastructure better than I do, but I'd be surprised if this is a big issue.

> And, finally, another small reason why I didn't like this change is because it makes the program not directly follow the developer's input. Right now, the width tolerance is very direct and easy to understand for developers - "If I set width_tolerance to 0.10, that means each width will increase by 10% every time", whereas with this change, they would no longer have control over how much the widths would change every time, which could be frustrating, and we don't want to take this clarity away from them.

I *seriously* doubt any developer is thinking this way. I don't think many developers care exactly what increment is between the candidate sizes, and I'd wager that many of those who do probably expect a linear increase in widths rather than an exponential one anyway. As I said above I'd expect `maxWidth` and `minWidth` are being set in the vast majority of cases anyway, which would mean per image they're getting a different range of candidate widths regardless.

But the option is called "width tolerance", not "precise scale factor". The algorithm both before and after this change works as advertised, and sticks to that given amount of tolerance. My change to the code causes the scale factor to stay the same or to shrink minimally, never increase, so the *tolerance* the developer requested is always still met.

The existing code does not follow the requested factor for each increment in most cases -- the last gap is usually smaller -- which is exactly what this change addresses.

I wonder if you'll reconsider this.

tremby:
The only reduced performance I can begin to see here is if more different images are being requested from your server as users resize their browser viewports or use different-sizes devices, after this change compared to before.

Yes, the gaps between candidate images will for the most part be slightly smaller, and so it's possible more different image sizes will be requested, assuming requests for the different candidate widths follow something resembling a bell curve.

But seeing that as "bad" requires the leap in logic that an 8% gap between image widths is acceptable (or perfect -- you did say it was very intentionally chosen) while slightly less than 8% is unacceptable. I really have to question this. If that is indeed the line of thinking, it seems to me the solution would be just to increase the default a little.

If my maths is right, the most the requested ratio can be reduced by my code is to cut it exactly in half (actually a mite under that due to floating point error), and this only happens in precisely the cases where the existing library's code would make exactly three candidates (where in my code `numSteps` would be calculated as 2), and where the last two are just 1 or 2 pixels apart in width. Example: with `minWidth` 615 and `maxWidth` 665 and 8% increases the old code gives 615, 664, 665. The changed code here would give 615, 640, 665. This is an edge case, and in my opinion the second sourceset is much more useful than the first.

This, by the way, is the most realistic example I could find of this edge case, and I brute-force searched the domain of min from 10 to 2000 and max from 20 to 8192 in increments of 5 pixels each, always with 8% requested step. This totals 567,138 possible ranges.

Taking into account only srcsets with 3 or more members (since 2 would heavily skew the results and would be the same before and after this code change anyway), over the entire range I brute-forced just now *including* the edge cases I mentioned above, the average calculated step was 7.7175%. Surely we can agree that this is not a big change from 8%.

Maybe the piece of the puzzle I'm missing is why 8% is special.

The code for my brute-force search:

```js
const ratios = [];
let minRatio = Infinity;
for (let min = 10; min <= 2000; min += 5) {
	for (let max = 20; max <= 8192; max += 5) {
		if (max <= min) continue;
		var numSteps = Math.ceil(Math.log(max / min) / Math.log(1.08));
		var adjustedIncrementRatio = Math.pow(max / min, 1 / numSteps);
		if (numSteps <= 1) continue;
		ratios.push(adjustedIncrementRatio);
		if (adjustedIncrementRatio <= 1.04) {
			console.log("min", min, "max", max, "steps", numSteps, "ratio", adjustedIncrementRatio);
		}
		if (adjustedIncrementRatio < minRatio) {
			minRatio = adjustedIncrementRatio;
		}
	}
}
console.log("avg over", ratios.length, "sets is", ratios.reduce((acc, r) => acc + r, 0) / ratios.length);
console.log("min is", minRatio);
```

sherwinski:
Hey @tremby 

I wanted to start by saying thanks again for the time and effort put into this PR and ensuing discussion. Our team is always excited to have these types of discussions with users, as it helps inform our understanding of how our SDK can be improved. Often times, part of this process requires that our team make choices about contributions based on the larger context that we have at an organizational level. We do our best to incorporate outside-contributions in each project, assuming they are aligned enough with that larger context, but can't promise everything will make it in. So to that end, I ask that you respect the project maintainers' decision not to merge this change and try to understand their position.

I also ask that you please kindly refrain from using a hostile tone when partaking in these discussions. As I said, this type of discourse is very important to us, but will only get so far if there is an aggressive atmosphere. Ultimately we are all working towards the same goal, so let’s try to be respectful even if there are differing ideas about what should/shouldn’t be implemented.

PR #206, opened 2021-01-14
feat: remove ensureEven requirement

ericdeansanchez:
The purpose of this PR is to remove the ensureEven requirement when
generating target widths. Prior, it was deemed necessary to restrict
target width values to be even-valued integers. Now, we've relaxed
this constraint, e.g. both `134` and `135` are acceptable target image
width values.


sherwinski (approved):
LGTM!
Can we also double check to make sure the default width list we're exporting here matches that of other projects that have removed the ensureEven constraint?

frederickfogerty:
For some context, Jason and I implemented this because we assumed that most devices had a even pixel width, and that most images at the small resolutions would be displayed full-bleed, leading to even pixel widths. Whether or not this reasoning was useful is up for debate.

frederickfogerty (commented):
LGTM!

PR #208, opened 2021-01-15
build(test): require esm for mocha testing, drop extra build steps

ericdeansanchez:


dist/imgix-core-js.cjs.js
```diff
@@ -396,15 +396,11 @@ var ImgixClient = /*#__PURE__*/function () {
         return this.targetWidthsCache[cacheKey];
       }
 
-      var ensureEven = function ensureEven(n) {
-        return 2 * Math.round(n / 2);
-      };
-
```
frederickfogerty, line 402: We don't usually include these changes in PRs

package.json
```diff
@@ -7,17 +7,18 @@
   "main": "dist/imgix-core-js.cjs.js",
   "module": "dist/imgix-core-js.esm.js",
   "browser": "dist/imgix-core-js.umd.js",
-  "type": "module",
```
frederickfogerty, line 10: Why was this removed? It seems necessary to support ESM for consumers.

frederickfogerty (changes requested):
LGTM, apart from the comments.

package.json
ericdeansanchez: It was removed because it caused conflicts with testing esm code with mocha. 

It seemed necessary to me as well, but it also seems to cause more trouble than what it's worth. Also we produce cjs, umd, and esm code here. Doesn't that support our ESM consumers?

If you know of a way of specifying `"type": "module"` in a way that doesn't conflict with our current setup and/or if specifying the type is _actually necessary_ lmk. I haven't found anything, but maybe I'm looking in the wrong places. Thanks!

sherwinski: What were the conflicts exactly?

ericdeansanchez: Mocha gets confused. It either wants `mjs` extensions everywhere and `"type": "module"` or `js` extensions (omitted i.a.w. @frederickfogerty recommendation) but without `"type": "module"` (and using the esm package).

IMO I prefer using the `.mjs` extensions explicitly (just seems a little cleaner/explicit). I made a PR for that solution as well. Right now I'm leaning towards PR #209 what do you think? @sherwinski

ericdeansanchez:
Closing this in favor of #209 will reopen if needed.

PR #209, opened 2021-01-15
feat: use mjs file extensions with type module

ericdeansanchez:


frederickfogerty (approved):
🔥🔥🔥

frederickfogerty:
@ericdeansanchez what has it been like using `.mjs` and the module system overall in the new version of Node? I've never used `.mjs` directly so I'm wondering what the experience was like

ericdeansanchez:
@frederickfogerty it's been smooth so far, tbh I haven't used them much either (just in this project). I also haven't seen a lot of projects using them explicitly (looked through more than a few `package.json` files from other projects haha).

PR #211, opened 2021-01-20
feat: enforce 0.01 lower bound for widthTolerance

ericdeansanchez:
The purpose of this PR is to enforce a lower bound on `widthTolerance`.
Prior, `widthTolerance` was expected to be greater than zero. However,
accepting a number like `0.000000001` here can do more harm than good.
We don't expect that such a number represents an actual use-case,
rather, we're enforcing this lower bound to keep defaults sensible and
help users guard against undesirable/accidental input.

Now, we enforce a lower bound of `0.01` on `widthTolerance`. This is also
pretty lenient, e.g. with the min/max width defaults set and setting the width
tolerance down to `0.01` (to one percent) we can make a srcset with 224
image URLs in it where the widths progress like `100, 102, 104, 106,
108, 110...`.

### DIFF
--- a/test/test-buildSrcSet.mjs
+++ b/test/test-buildSrcSet.mjs
@@ -1079,9 +1079,20 @@ describe('SrcSet Builder:', function describeSuite() {
             domain: 'testing.imgix.net',
           }).buildSrcSet('image.jpg', {}, { widthTolerance: 999999.999 });
 
-          const actualLength = srcset.split(',').length;
+          const srcsetSplit = srcset.split(',');
+          const actualLength = srcsetSplit.length;
+
+          const srcsetMin = Number.parseFloat(
+            srcsetSplit[0].split(' ')[1].slice(0, -1),
+          );
+
+          const srcsetMax = Number.parseFloat(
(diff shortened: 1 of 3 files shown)


test/test-buildSrcSet.mjs
```diff
@@ -1074,6 +1074,16 @@ describe('SrcSet Builder:', function describeSuite() {
...
         });
 
+        it('produces srcset with min and max widths when widthTolerance is large', function testSpec() {
+          const srcset = new ImgixClient({
+            domain: 'testing.imgix.net',
+          }).buildSrcSet('image.jpg', {}, { widthTolerance: 999999.999 });
+
+          const actualLength = srcset.split(',').length;
+
+          assert.strictEqual(actualLength, 2);
```
sherwinski, line 1084: Can we include a test ensuring that these values are the min and max?

sherwinski (changes requested):

frederickfogerty (approved):
LGTM ✌️ I'm wondering if it's worthwhile also bounding `widthTolerance` to be at max `1` as well

sherwinski (approved):

PR #212, opened 2021-01-21
test: const where possible, fix assert order

ericdeansanchez:
The purpose of this PR is to ensure we use `const` where possible as
well as use the correct assertion order (i.e. actual, then expected).
Prior, there were some leftover `var`s etc. Now those have been removed
from multiple files.

Now we use `strictEqual` and `notStrictEqual`; we also use the assert
order expected by the assert functions.


sherwinski (approved):

frederickfogerty (approved):
Nice 👍

PR #214, opened 2021-01-21
test: deduplicate code using functions

ericdeansanchez:
The purpose of this PR is to wrap duplicated code into functions.
This PR also fixes a bug in the testing of width tolerance increases.
Prior, we were testing that a 40% increase (`1.40`) was less than the
desired bound of `2.40`. Now, we add a small amount to the tolerance
upper bound (`0.004`) so that when we check increases that are slightly
more than 40% (e.g. 40.1%) we account for the rounding that occurs in
generating target widths.

### DIFF
--- a/test/test-buildSrcSet.mjs
+++ b/test/test-buildSrcSet.mjs
@@ -72,7 +72,7 @@ function assertIncludesQualities(srcset, qualities) {
 }
 
 function assertIncludesQualityOverride(srcset, qOverride) {
-  const _ = srcset.split(',').map((u) => {
+  srcset.split(',').map((u) => {
     const url = u.split(' ')[0];
     assert(url.includes(`q=${qOverride}`));
   });
--- a/test/test-buildSrcSet.mjs
+++ b/test/test-buildSrcSet.mjs
@@ -58,7 +58,7 @@ function assertMinMaxWidthBounds(srcset, minBound, maxBound) {
 
 function assertCorrectWidthDescriptors(srcset, descriptors) {
   const srcsetSplit = srcset.split(',');
-  const _ = srcsetSplit.map((u, i) => {
+  srcsetSplit.map((u, i) => {
     const width = parseInt(u.split(' ')[1].slice(0, -1), 10);
     assert.strictEqual(width, descriptors[i]);
   });
--- a/test/test-buildSrcSet.mjs
+++ b/test/test-buildSrcSet.mjs
@@ -65,7 +65,7 @@ function assertCorrectWidthDescriptors(srcset, descriptors) {
 }
 
 function assertIncludesQualities(srcset, qualities) {
-  const _ = srcset.split(',').map((u, i) => {
+  srcset.split(',').map((u, i) => {
     const url = u.split(' ')[0];
     assert(url.includes(`q=${qualities[i]}`));
   });
--- a/test/test-buildSrcSet.mjs
+++ b/test/test-buildSrcSet.mjs
@@ -2,64 +2,166 @@ import md5 from 'md5';
 import assert from 'assert';
 import ImgixClient from '../src/main.mjs';
 
+function assertWidthsIncreaseByTolerance(srcset, tolerance) {
+  const srcsetWidths = srcset.split(',').map((u) => {
+    const tail = u.split(' ')[1];
+    const width = tail.slice(0, -1);
+    return Number.parseFloat(width);
+  });
+
+  // Make two equal sized arrays one for the numerators, e.g.
+  // [x1, x2, ..., xN] and another for the denominators, e.g.
+  // [x0, x1,..., x(N-1)].
+  const numerators = srcsetWidths.slice(1);
+  const denominators = srcsetWidths.slice(0, -1);
+
+  // Zip the numerator/denominator pairs.
+  const pairs = numerators.map((n, i) => {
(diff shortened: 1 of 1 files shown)


test/test-buildSrcSet.mjs
```diff
@@ -2,489 +2,348 @@ import md5 from 'md5';
...
+  const max = Number.parseFloat(
+    srcsetSplit[srcsetSplit.length - 1].split(' ')[1].slice(0, -1),
+  );
+  assert(min >= minBound);
+  assert(max <= maxBound);
+}
+
+function assertCorrectWidthDescriptors(srcset, descriptors) {
+  const srcsetSplit = srcset.split(',');
+  const _ = srcsetSplit.map((u, i) => {
```
frederickfogerty, line 61:
```suggestion
  srcsetSplit.map((u, i) => {
```

test/test-buildSrcSet.mjs
```diff
@@ -2,489 +2,348 @@ import md5 from 'md5';
...
+    const width = parseInt(u.split(' ')[1].slice(0, -1), 10);
+    assert.strictEqual(width, descriptors[i]);
+  });
+}
+
+function assertIncludesQualities(srcset, qualities) {
+  const _ = srcset.split(',').map((u, i) => {
```
frederickfogerty, line 68:
```suggestion
  srcset.split(',').map((u, i) => {
```

test/test-buildSrcSet.mjs
```diff
@@ -2,489 +2,348 @@ import md5 from 'md5';
...
+    const url = u.split(' ')[0];
+    assert(url.includes(`q=${qualities[i]}`));
+  });
+}
+
+function assertIncludesQualityOverride(srcset, qOverride) {
+  const _ = srcset.split(',').map((u) => {
```
frederickfogerty, line 75:
```suggestion
  srcset.split(',').map((u) => {
```

test/test-buildSrcSet.mjs
```diff
@@ -2,489 +2,348 @@ import md5 from 'md5';
...
+  lastParts.map((d, i) => {
+    // assert 1x === `${0 + 1}x`, etc.
+    assert.strictEqual(d, `${i + 1}x`);
+  });
+}
+
+function assertDoesNotIncludeQuality(srcset) {
+  const _ = srcset.split(',').map((u) => {
+    const url = u.split(' ')[0];
+    assert(!url.includes(`q=`));
```
frederickfogerty, line 106: nit: this might fail if there's any substring in the URL which is `q=`. I'm imagining there might be a param which ends in q which would trigger this.

```suggestion
    assert(!Array.from((new URL(url)).searchParams.keys()).includes("q"));
```

You might consider using the `URL` class elsewhere as well, it makes working with URLs much easier

frederickfogerty (approved):
Nice 👏

test/test-buildSrcSet.mjs
```diff
@@ -2,489 +2,348 @@ import md5 from 'md5';
 import assert from 'assert';
 import ImgixClient from '../src/main.mjs';
 
+function assertWidthsIncreaseByTolerance(srcset, tolerance) {
+  const srcsetWidths = srcset.split(',').map((u) => {
+    const tail = u.split(' ')[1];
+    const width = tail.slice(0, -1);
```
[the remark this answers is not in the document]
ericdeansanchez: Oh nice!

test/test-buildSrcSet.mjs
```diff
@@ -2,489 +2,348 @@ import md5 from 'md5';
...
+  const numerators = srcsetWidths.slice(1);
+  const denominators = srcsetWidths.slice(0, -1);
+
+  // Zip the numerator/denominator pairs.
+  const pairs = numerators.map((n, i) => {
+    return [n, denominators[i]];
+  });
+
+  // Be as tolerant as we can.
+  const tolerancePlus = tolerance + 0.004;
```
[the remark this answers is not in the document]
ericdeansanchez: Say we're trying to ensure that each subsequent width is 40% larger than the last. We're doing floating point arithmetic so we have to be careful when doing comparisons. In the test where we're ensuring 40% increases we get these ratios:

```
0.40145...
0.40104...
```

If these were convented to percentages, i.e. `40.1%`, and then rounded they'd still be `40%`. The last number to round down to 40% would be `40.4%` (whereas `Math.round(40.5) == 41` which would mean we're checking for 41% instead of 40%).

We could loosen it a bit more if we want to. I just noticed that in our test where we were checking for `0.40` increases the bound we were checking against was actually `2.4`. This might have happened because 1.40104 is not less than 1.40 but is less than 2.4.

test/test-buildSrcSet.mjs
ericdeansanchez: Habit. I wrote it like this to make it explicit that this function:

- does not produce a result  and
- the call to `map` doesn't intend to produce a consumable result

Tbh I wasn't sure if this went against idioms

test/test-buildSrcSet.mjs
ericdeansanchez: Totally agree. I've been cooking up a project that employs more use of `URL` and other utilities to do testing. For this case here I'm going to make a note of this and then work it in.

PR #215, opened 2021-01-22
feat: create a DPR srcset when a fixed height is specified

### LINKED ISSUE #177: A fluid-width srcset is generated when a height is passed

**Describe the bug**
A fluid-width (i.e. `srcset="url 100w, url2 116w, ..."`) srcset is returned even when a fixed height is set. This is probably not the right behaviour, since this results in the image changing dimensions/aspect ratio as the width of the `img` or `picture` element changes.

**To Reproduce**
```js
const client = new ImgixClient({ domain: 'assets.imgix.net' });
client.buildSrcSet('/image.jpg', { h: 100 });
```

**Expected behavior**
A "DPR" srcset is generated.

**Information:**
 - imgix-core-js version: 2.3.2

ericdeansanchez:
The purpose of this PR is to ensure we create a DPR-based srcset when
a fixed height, `h`, is specified. Prior, we required that a `h` and
an aspect ratio (`ar`) be passed. Now, we require that only  a `w` or
`h` be specified to trigger a DPR-based srcset to be created.

Closes #177

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ https://testing.imgix.net/image.jpg?w=8192&s=59eb881b618fed314fe30cf9e3ec7b00 81
 
 #### Fixed Image Rendering
 
-In cases where enough information is provided about an image's dimensions, `buildSrcSet()` will instead build a `srcset` that will allow for an image to be served at different resolutions. The parameters taken into consideration when determining if an image is fixed-width are `w`, `h`, and `ar`. By invoking `buildSrcSet()` with either a width **or** the height and aspect ratio (along with `fit=crop`, typically) provided, a different `srcset` will be generated for a fixed-size image instead.
+Specifying either a `w` or a `h` parameter to `buildSrcSet()` will create a DPR-based srcset. This DPR-based srcset allows for the fixed-sized image to be served at different resolutions (i.e. at different pixel densities).
 
 ```js
 var client = new ImgixClient({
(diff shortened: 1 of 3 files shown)


sherwinski (approved):
LGTM 👍

src/main.mjs
```diff
@@ -103,9 +103,9 @@ export default class ImgixClient {
   }
 
   buildSrcSet(path, params = {}, options = {}) {
-    const { w, h, ar } = params;
+    const { w, h } = params;
 
-    if (w || (h && ar)) {
+    if (w || h) {
```
frederickfogerty, line 108: Little nit that a truthy check should never be used with numbers. Instead, this should be:

```js
if (w != null || h != null) {
```

README.md
```diff
@@ -180,7 +180,7 @@ https://testing.imgix.net/image.jpg?w=8192&s=59eb881b618fed314fe30cf9e3ec7b00 81
 
 #### Fixed Image Rendering
 
-In cases where enough information is provided about an image's dimensions, `buildSrcSet()` will instead build a `srcset` that will allow for an image to be served at different resolutions. The parameters taken into consideration when determining if an image is fixed-width are `w`, `h`, and `ar`. By invoking `buildSrcSet()` with either a width **or** the height and aspect ratio (along with `fit=crop`, typically) provided, a different `srcset` will be generated for a fixed-size image instead.
+Specifying either a `w` or a `h` parameter to `buildSrcSet()` will create a DPR-based srcset. This DPR-based srcset allows for the fixed-sized image to be served at different resolutions (i.e. at different pixel densities).
```
frederickfogerty, line 183: Maybe better to even just say "for the fixed-sized image to be served for different pixel densities."

PR #216, opened 2021-01-22
refactor: improve target widths generation

ericdeansanchez:
Prior, this function performed an array access, property access, and
a comparison in the loop condition. Now, this has been reduced to a
single comparison. Prior, this function made an additional function
call/comparison in the body (i.e. `Math.min`) on every iteration. Now,
that has been removed as well.

Prior, this function attempted to return a list resolutions before
checking the cache (i.e. `[minW]`). Now, we check the cache first.

Now, when the loop terminates we know that the last width to be pushed
is less than the currentWidth that caused the loop condition to fail
and its value is greater than or equal to `maxW`. We want to stop at
`maxW` exactly so we push it if it's less than the last resolution.

### DIFF
--- a/src/main.mjs
+++ b/src/main.mjs
@@ -165,25 +165,35 @@ export default class ImgixClient {
 
   // returns an array of width values used during srcset generation
   _generateTargetWidths(widthTolerance, minWidth, maxWidth) {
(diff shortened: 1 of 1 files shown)


sherwinski (approved):

frederickfogerty (commented):
LGTM

PR #218, opened 2021-01-26
docs(readme): const where possible

ericdeansanchez:
The purpose of this PR is to update the readme to use `const` where possible.


frederickfogerty (approved):
Loving these `const`ant changes 👏

sherwinski (approved):

PR #219, opened 2021-01-26
refactor(perf): ensure prettier works with benchmarks

ericdeansanchez:
Prior, prettier didn't take test/performance into account. Now, it does.
The file extension of the prettier config was changed to `.cjs` to
ensure compatibility with the CLI and our package.json specification of
`"type": "module"`.

Prior, our prettier setup provided editor support (i.e. format on save),
but running the script (i.e. `npm run pretty`) would error on the benchmark
suite. Now, that has been fixed and both work as expected.


frederickfogerty (approved):
Great stuff 👏

sherwinski (approved):

PR #222, opened 2021-01-27
feat: drop bower.json

ericdeansanchez:


sherwinski (approved):

frederickfogerty:
Pedantic nit, but this probably isn't a `feat`, and is probably a `chore`

PR #226, opened 2021-02-02
feat: update asser_version script

ericdeansanchez:
This PR updates the `assert_version.js` script to have the same functionality
as before with the only exception that it's pulling the version, from `constants.mjs`.

In a future PR I think we'll add semantic release here so this may change again. This
PR is meant to ensure this repo mimics the previous version 1:1.


sherwinski (approved):

frederickfogerty (commented):
LGTM!

PR #239, opened 2021-02-22
chore(rename): org scope package

ericdeansanchez:
The purpose of this PR is to start the renaming process. `dist/` files have been
included to make the subsequent steps a little cleaner (less noise for the next
reviewal).

If/when this PR gets a thumbs up, I'll start the next steps:

- do repo/project renaming
- release beta
- use beta to ensure the readme is accurate and updated
- then do the changelog edits

Lmk if you think there's a better strategy/order of steps.


src/index.mjs
```diff
@@ -34,7 +34,7 @@ export default class ImgixClient {
     }
 
     if (this.settings.includeLibraryParam) {
-      this.settings.libraryParam = 'js-' + VERSION;
+      this.settings.libraryParam = 'js-core-' + VERSION;
```
sherwinski, line 37: Is this necessary if the version number is staying on track with what we currently have?

sherwinski (approved):
LGTM 👍 

I noticed this and was wondering if git was feeling ok:

Screen Shot 2021-02-22 at 5 04 47 PM

Screen Shot 2021-02-22 at 5 05 12 PM

frederickfogerty:
I'm still going to review this at a later stage but my 2c is that it'd be really nice to fix up the renames that @sherwinski mentioned so that we keep git history for the future.

src/index.mjs
ericdeansanchez: No it's not necessary, but I had wondered what you thought about it. I'm good with keeping just the `js-` portion.

assert_version.cjs
```diff
@@ -6,7 +6,7 @@ var packageVersion = require('./package').version;
     return 0;
   } else {
     process.stdout.write(
-      'FAIL: package.json and lib/imgix-core-js.js versions do not match!\n',
+      'FAIL: package.json and @imgix/js-core versions do not match!\n',
```
frederickfogerty, line 9:
```suggestion
      'FAIL: package.json and src/constants.mjs versions do not match!\n',
```

dist/imgix-js-core.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for imgix-core-js.js
+// Type definitions for @imgix/js-core.js
```
frederickfogerty, line 1:
```suggestion
// Type definitions for imgix-js-core.*.js
```

I would also accept just `@imgix/js-core` without the `.js`

frederickfogerty (changes requested):
Some places I found `imgix-core-js` still existing:

- `.prettierignore`
- `CHANGELOG.md` (you said you were going to handle this later so that's fine)
- `package-lock.json` (since package.json has changed we need to change this)
- `README.md`
- `.github` folder (PR template, etc)

Also, if we could somehow ensure the git renames work as expected to keep history that would be amazing. Sometimes it is not possible and needs to be done with two commits/PRs.

dist/imgix-js-core.d.ts
ericdeansanchez: Nice catch!!!

ericdeansanchez:
TODO:
- [ ] `.prettierignore`
- [ ] `CHANGELOG.md` (you said you were going to handle this later so that's fine)
- [x] `package-lock.json` (since package.json has changed we need to change this)
- [ ] `README.md`
- [ ] `.github` folder (PR template, etc)

ericdeansanchez:
@frederickfogerty we split the commits but the diff on GitHub still displays the same weirdness that @sherwinski posted above. Locally, git displays the correct thing (both pre and post rebasing/re-managing the commits).

PR #248, opened 2021-03-04
feat: static targetWidths functionality

ericdeansanchez:
The purpose of this PR is to extend `targetWidths` functionality to
users via a static class function. A static function has been used to
preserve the import syntax for UMD, CommonJS, and ES6 modules.

Previously, this function was named `_generateTargetWidths` which
required users to instantiate an `ImgixClient` object to use it. Now,
this function has been renamed and re-implemented as a static function.

Previously, the parameter order required `widthTolerance`, `minWidth`,
and then `maxWidth`. The order has been changed in accordance with
how common a particular argument is expected to be passed:

```js
targetWidths(minWidth, maxWidth, widthTolerance, cache)
```

### DIFF
--- a/src/index.mjs
+++ b/src/index.mjs
@@ -112,26 +112,70 @@ export default class ImgixClient {
     }
   }
 
+  // returns an array of width values used during srcset generation
+  static targetWidths(
+    minWidth = 100,
+    maxWidth = 8192,
+    widthTolerance = 0.08,
+    cache = {},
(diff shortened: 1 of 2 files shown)


sherwinski (approved):

frederickfogerty:
Thanks so much for this @ericdeansanchez! Could we please update the TS definitions to include this change?

PR #252, opened 2021-03-08
chore(types): include targetWidths

ericdeansanchez:
The purpose of this PR is to include the `targetWidths` function
in the `.d.ts` file.


sherwinski (approved):
LGTM

src/imgix-js-core.d.ts
```diff
@@ -15,7 +15,7 @@ declare class ImgixClient {
     buildSrcSet(path: string, params?: {}, options?: SrcSetOptions): string;
     _buildSrcSetPairs(path: string, params?: {}, options?: SrcSetOptions): string;
     _buildDPRSrcSet(path: string, params?: {}, options?: SrcSetOptions): string;
-
+    targetWidths(minWidth?: number, maxWidth?: number, widthTolerance?: number, cache?: {}): number[];
```
frederickfogerty, line 18: IIRC targetWidths is static so this should be 
```ts
    static targetWidths(...)
```

PR #257, opened 2021-03-10
fix: validate minWidth, maxWidth, widthTolerance

ericdeansanchez:


sherwinski (approved):

test/test-buildSrcSet.mjs
```diff
@@ -763,11 +763,29 @@ describe('SrcSet Builder:', function describeSuite() {
...
+        it('produces the default target width resolutions given default args', function testSpec() {
           const actual = ImgixClient.targetWidths();
           assert.deepStrictEqual(actual, RESOLUTIONS);
         });
+
+        it('errors on invalid minWidth', function testSpec() {
+          assert.throws(() => {
+            ImgixClient.targetWidths(0);
+          }, Error);
+        });
```
frederickfogerty, line 776: I think we could a liiitle better with these tests, to make sure we test other boundary conditions.

For `minWidth`:

- just below 1, so `0.99`
- zero (already tested)
- negative number, e.g. `-0.5`

For `widthTolerance`:

- just below threshold, so `0.0099`
- zero, `0`
- negative number, e.g. `-1`

frederickfogerty (approved):
Looks pretty good! Tests could be improved a little bit but looks good!

test/test-buildSrcSet.mjs
ericdeansanchez: Sure thing!

PR #259, opened 2021-03-16
build: drop module type from package.json

ericdeansanchez:


frederickfogerty (commented):
I'll reply more tomorrow, but I'm not sure this is the _best_ solution. I think maybe we could change the `browser` field to not point to the UMD bundle as well 🤷

ericdeansanchez:
@frederickfogerty Okay, cool I'll hold off.

ericdeansanchez:
Considering closing this in favor of #260

frederickfogerty:
Yeah I think we should close this 👍

PR #260, opened 2021-03-18
fix: remove type and browser attributes from package.json

ericdeansanchez:


luqven (commented):
LGTM 👍🏼 , but will leave it to Fred and / or Sherwin to approves since I'm not as familiar with how removing `"browser"` will affect things downstream.

frederickfogerty (approved):
Yeah let's do it 👍

PR #267, opened 2021-03-24
fix: improve error messages for target width validator

frederickfogerty:
The error message isn't quite correct, so I improved it a little.


ericdeansanchez (approved):
Yes!!!

frederickfogerty:
I've rebased this onto `main` after the `.js` changes

PR #282, opened 2021-05-03
test(encoding): add path encoding tests to img url that resolves

luqven:
# Description

This PR adds more path encoding tests.

## Up for discussion

This PR chose to make a new file for these tests. An argument could be make for leaving these inside the `buildURL` tests. For the sake of making these as easy to read and find as possible, they were left in a separate file.

### DIFF
--- a/test/test-pathEncoding.js
+++ b/test/test-pathEncoding.js
@@ -1,0 +1,39 @@
+import assert from 'assert';
+import ImgixClient from '../src/index.js';
+
+describe('Path Encoding:', function describeSuite() {
+  // NOTE: the image urls tested bellow actually resolve to an image.
+  describe('buildURL', function describeSuite() {
+    let client;
+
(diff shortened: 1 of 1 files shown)


sherwinski (approved):
Overall looks good but I'm a little hesitant about using the `sherwinski` source here. IMO we should use `sdk-test`/`sdk-proxy-test` everywhere possible and make sure the source serves live images we can actually test against.

luqven:
> Overall looks good but I'm a little hesitant about using the `sherwinski` source here. IMO we should use `sdk-test`/`sdk-proxy-test` everywhere possible and make sure the source serves live images we can actually test against.

Forced pushed to remove the checkin in domain and replace it with `sdk-test.imgix.net`.

sherwinski:
Perfect 🎉 thanks @luqven!

PR #286, opened 2021-06-03
fix: ensure undefined parameters not added to URL 

luqven:
This PR ensures that undefined parameters are not added to the
params object generated by _buildParams. This way, undefined params do
not then end up in the generated URL.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -67,14 +67,18 @@ export default class ImgixClient {
         : []),
 
       // Map over the key-value pairs in params while applying applicable encoding.
-      ...Object.entries(params).map(([key, value]) => {
+      ...Object.entries(params).reduce((prev, [key, value]) => {
+        if (value == null) {
+          return prev;
+        }
         const encodedKey = encodeURIComponent(key);
         const encodedValue =
           key.substr(-2) === '64'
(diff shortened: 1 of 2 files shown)


sherwinski (commented):
For my context, would you mind explaining what the motivation for this PR is? In isolation, it _seems_ fine but I'm hesitant to add guardrails for scenarios that don't sound likely to happen?

That is, unless this is proactively guarding against something that we know _is happening_, such as in a consuming library.

frederickfogerty:
@sherwinski there was a bug occurring in Gatsby that was caused by this. Basically, we had some code like this

```js
const imgixParams = {
  ...args.imgixParams,
  ar: (width != null && height != null) ? `${width/height}:1` : undefined
};

const src = client.buildURL(path, imgixParams); // => contains ar=undefined
```

It's a pretty common pattern that if an object value is `undefined` or `null` it should be treated as if it is not set. To workaround this we would have to manually prune all `null` and `undefined` from the `imgixParams` object before passing it to `js-core`.

frederickfogerty (approved):
✅

sherwinski:
@frederickfogerty I guessed as much, but figured I ask anyways.

PR #287, opened 2021-06-04
(PoC) feat: add _buildURL static method

luqven:
> This PR is a WIP
> This PR is a stacked PR

| PR 	| Title 	| Merges Into 	|   diff	|
|----	|-------	|-------------	|---	|
|    1	|    [_buildURL](https://github.com/imgix/js-core/pull/287) 👈🏼 	|           main     	|   [review this](https://github.com/imgix/js-core/pull/287/files)	|
|    2	|        [_buildSrcSet](https://github.com/imgix/js-core/pull/288)   	|           #287    	|   [review this](https://github.com/imgix/js-core/pull/288/files/2bddedfb20218fb975847427185ea30f3aac1a54..80806b276c58dc120784903ec5ac913e1bc311eb)	
|    3	|    [fix extractUrl](https://github.com/imgix/js-core/pull/290)  	|           #288     	|   [review this](https://github.com/imgix/js-core/pull/290/files/80806b276c58dc120784903ec5ac913e1bc311eb..096d101)	|

This PR creates the `_buildURL` static method on the ImgixClient. This methods allows us to use a 1-step, or full, URL to generate imgix formatted urls.

## Future work
Proxy URL tests are needed here. They're not being handled as of yet.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -75,19 +75,7 @@ export default class ImgixClient {
 
     const client = new ImgixClient({ domain, ...options });
 
-    if (options.sanitize) {
-      return client.buildURL(path, params);
-    } else {
-      // build the params string
-      let formattedParams = client._buildParams(params);
-      // sign the params string
-      if (!!client.settings.secureURLToken) {
-        formattedParams = client._signParams(path, formattedParams);
-      }
-
-      // return the url + params
-      return prefix + url + formattedParams;
-    }
+    return client.buildURL(path, params);
   }
 
   _buildParams(params = {}) {
--- a/src/index.js
+++ b/src/index.js
@@ -16,6 +16,8 @@ import {
   validateWidthTolerance,
 } from './validators.js';
 
+import { extractUrl } from './helpers';
+
 export default class ImgixClient {
   constructor(opts = {}) {
     this.settings = { ...DEFAULT_OPTIONS, ...opts };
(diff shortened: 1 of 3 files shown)


frederickfogerty (commented):
I think this is a good start but I had a few points/questions:

1. Is `ImgixClient._buildOneStepURL` the intended public API for this? If so, I don't there should be a `_` in the public API, and I think we can potentially come up with a better name.
2. I'm curious about this `sanitize` option - when would we `not` want the URL to be sanitized?
3. Although this is a "closed" change (i.e. doesn't affect the original source code), I can't help but think if there is maybe a better way to structure this so it's not necessary to split the URL and then join it back together again. i.e. make `extractUrl` not necessary
4. Proxy URLs???

src/index.js
```diff
@@ -59,6 +61,35 @@ export default class ImgixClient {
     );
   }
 
+  static _buildOneStepURL({ url, params, opts }) {
```
sherwinski, line 64: I'm curious what the specific motivation is for accepting these parameters as an object? It seems a little unusual since none of the other instance methods expect their parameters this way.

sherwinski (commented):
> I think we can potentially come up with a better name.

What do we think about having the name mirror the class method, ie `ImgixClient._buildURL`?

> I'm curious about this sanitize option - when would we not want the URL to be sanitized?

For the most part I agree that we should always sanitize, but I think we need to first think through the proxy use case a bit more. How do we expect someone to pass a proxy URL through this API? ie `https://proxy.imgix.net/https://assets.imgix.net/pione.jpg?s=...`? Frankly, that seems like an awkward experience and I don't think we should expect users to pass it in pre-encoded.

Initially, I felt that users should stick to using the instance methods if they want to build proxy URLs but I'm willing to change my mind.

src/index.js
luqven: Ah that's a habit of mine more than anything. 

I've gotten object parameters drilled into me from past experience. Purported benefits is that you can never get the order of the arguments wrong and you create fewer breaking changes down the road when you add and or remove parameters. But more than happy to remove this for consistency's sake and stick to normal parameters if that's perferable.

luqven:
> Initially, I felt that users should stick to using the instance methods if they want to build proxy URLs 

I second this ☝🏼 . `sanitize` option can and IMO should be removed.

As for proxies, for one-step URLs we can treat proxies as an `opts` parameter. Something like

```js
const url = "https://foo.bar.net"
const opts = { proxy: "https://bar.baz.net/img.jpg" }
const params = {fit: "crop"}
_buildOneStepURL({ url, params, opts }) {

// inside _buildOneStepURL, we'd call extractUrl where...
export function extractUrl({ url, opts }) {  
//... we can add the proxy path
  path = _path.join('/');
  if (options.proxy) {
    path = client._sanitizePath(options.proxy);
  }
  return {  prefix,  domain,  path,  };
}
```

sherwinski:
@luqven Yeah that sounds like a plausible solution. Since this will probably take a few beta releases to get right, I'd say feel free to experiment with the interface until all the wrinkles get ironed out 👍

luqven:
Updated this PR to rename the static method to `_buildURL`, rename `opts` to `options`, and rename the test suite to match.

luqven:
Closing in favor of #294

PR #289, opened 2021-06-04
fix: remove "v" prefix from `VERSION` constant

sherwinski:
The inclusion of the prefix (e.g. `ixlib=js-v3.1.3`) causes issues in our internal reporting that result in versions of this libray with the prefix to not be correctly captured. This commit removes the prefix to match the format previous to the current major version.


frederickfogerty (approved):
LGTM ✅

PR #293, opened 2021-06-22
build: remove stale .d.ts file from dist

sherwinski:
Noticed there was an old version of the project's TS definitions file floating around in the dist directory.


luqven:
Ah that's my bad. I saw this and figured for some reason the OG was missing. Didn't realize this was an out dated version.

luqven (approved):
LGTM

PR #310, opened 2022-01-07
chore(dev-deps): replace dtslint with tsd

sherwinski:
This PR replaces the type-checking library dtslint with tsd in this project. The motivation for this was to replace some outdated transitive dependencies to dtslint, as well as use a [more accepted](https://www.npmtrends.com/dtslint-vs-tsd-vs-typings) testing library.


frederickfogerty (approved):
LGTM. I think the changes to `src/constants.js`, `src/index.js`, and `test/test-buildSrcSet.js` should be removed from this, otherwise this should be good to go.

PR #314, opened 2022-02-17
feat: allow path encoding to be disabled

frederickfogerty:
Before this PR, this library would encode all paths passed to either `buildURL` or `buildSrcSet`. This was a great default in a lot of scenarios, but this proved to be irritating when integrating with some data sources that provided already encoded URL paths. Thus, this PR allows this path encoding behavior to be disabled in the `options` parameter for both methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,14 @@
     * [In-browser](#in-browser)
 - [Configuration](#configuration)
 - [API](#api)
-    * [`ImgixClient.buildURL(path, params)`](#imgixclientbuildurlpath-params)
+    * [`ImgixClient.buildURL(path, params, options)`](#imgixclientbuildurlpath-params-options)
     * [`ImgixClient.buildSrcSet(path, params, options)`](#imgixclientbuildsrcsetpath-params-options)
         + [Fixed Image Rendering](#fixed-image-rendering)
         + [Custom Widths](#custom-widths)
         + [Width Tolerance](#width-tolerance)
         + [Minimum and Maximum Width Ranges](#minimum-and-maximum-width-ranges)
         + [Variable Qualities](#variable-qualities)
+        + [Disable Path Encoding](#disable-path-encoding)
     * [Web Proxy Sources](#web-proxy-sources)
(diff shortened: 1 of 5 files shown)


frederickfogerty:
Current dependencies on/for this PR:
* main
  * **PR #314** <a href="https://app.graphite.dev/github/pr/imgix/js-core/314" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://graphite.dev?utm_source=gh).

sherwinski:
@frederickfogerty I just realized that this change is probably not compatible with URL signing as it relies on the path being sanitized/encoded first.

frederickfogerty:
@sherwinski I made the trailing slash fix/update here, btw

sherwinski:
@frederickfogerty I haven't had chance to dig in more but I should have more time today to build out an example that confirms/denies my suspicion. I'll follow up

sherwinski:
> Does this mean that URL signing can't be used with already encoded URLs?

Yeah I'm saying it won't work unless the URL is encoded exactly as we would during the sanitizing step. To me that just seems like a bit of smell since from a user's perspective, it might not be that apparent that path encoding and URL signing are that tightly coupled.

For example, this test will fail (the `expected` value includes the signature if path encoding were true):
```js
const client = new ImgixClient({
  domain: 'sdk-test.imgix.net',
  includeLibraryParam: false,
  secureURLToken: 'abc123',
});

const actual = client.buildURL(
  '/file+with%20some+crazy?things.jpg',
  {
    w: 100,
  },
  {
    disablePathEncoding: true,
  },
);

const expected =
  'https://sdk-test.imgix.net/file%2Bwith%2520some%2Bcrazy%3Fthings.jpg?w=100&s=0c4068b0df31a8a8b67dd79247063939';

assert.strictEqual(actual, expected);
```

Not sure if the best path here is to document this as a caveat or throw a warning if users attempt to sign a URL while also disabling the encoding step.

sherwinski (approved):

PR #316, opened 2022-03-02
feat: add _buildURL and _buildSrcSet static methods

luqven:
This PR creates static `_buildURL` and `_buildSrcSet` methods on the imgix client. These methods allow absolute URLs to be formatted for use with the imgix client.

- _buildURL

```
   *`_buildURL` static method allows full URLs to be formatted for use with
   * imgix.
   *
   * - If the source URL has included parameters, they are merged with
   * the `params` passed in as an argument.
   * - URL must match `{host}/{pathname}?{query}` otherwise an error is thrown.
```

- _buildSrcSet:    

```
   * _buildSrcSet static method allows full URLs to be used when generating
   * imgix formatted `srcset` string values.
   *
   * - If the source URL has included parameters, they are merged with
   * the `params` passed in as an argument.
   * - URL must match `{host}/{pathname}?{query}` otherwise an error is thrown.
  ```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,17 @@
 {
   "name": "@imgix/js-core",
-  "version": "3.4.0",
+  "version": "3.5.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@imgix/js-core",
-      "version": "3.4.0",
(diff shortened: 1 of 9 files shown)


sherwinski (approved):
LGTM

PR #328, opened 2022-09-13
chore(types): remove top level options

luqven:
This commit removed incorrect type declarations that exposed top-level settings as client properties. These options should only be used inside the constructor or as function arguments for static methods.


luqven:
Closing in favor of #342

PR #329, opened 2022-09-14
chore: add gh issue action

luqven:
Add GH action that adds an issue with a `bug` or `needs-triage` label to the issue board.


atlawrie (approved):
Discussed offline, LGTM.

PR #330, opened 2022-09-16
ci(build): test if builds are successful during ci test step

atlawrie:
Adds an additional test step to our CircleCI process to test successful builds.

# Before
Problems with builds can only be detected by CircleCI during the deploy step; this step only runs on `main` and prerelease branches. As a result, commits that break builds (included automated dependency updates by Renovate) in many cases might not be detected until after merge.

# After
We now test for a successful build step on every branch, using both stable and LTS node versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,12 @@ jobs:
       - checkout
       - node/install-packages
       - run: npm run test
+      - run:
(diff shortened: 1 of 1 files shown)


luqven (approved):
LGTM 👍🏼 

Let's hold off on merging until next week to avoid a Friday release 🤣 .

PR #346, opened 2022-12-02
[SUPPORT-445] JS-CORE-updates

AbuBoo:


AbuBoo:
Updating JS-core documentation page on install. 
[https://docs.imgix.com/libraries/js-core](url)

Updating documentation to include variable definition in the needed code block  

Line below being added to code block in image. 
`const ImgixClient = require('@imgix/js-core');`


Screen Shot 2022-12-02 at 3 48 36 PM


this is done to make sure the install works right

luqven:
Resolved offline

PR #350, opened 2023-01-12
fix: apply custom encoder to base64 parameters

sherwinski:
## Description

Before this PR, the `encoder` buildParams option would only be applied to the non-base64 parameters (i.e. parameters whose values did not end in `64`). After this PR, the same encoder will be applied to base64 parameters as well.

It also adds an optional parameter `key` to the encoder which allows users to specify encoding logic conditional to the parameter key. This is helpful if users wish to encode base64 and non-base64 parameters differently.

## Checklist

- [x] Read the [contributing guidelines](CONTRIBUTING.md).
- [x] Each commit follows the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) spec format.
- [x] Update the readme (if applicable).
- [] ~~Update or add any necessary API documentation (if applicable)~~
- [x] All existing unit tests are still passing (if applicable).
- [] ~~Add some [steps](#steps-to-test) so we can test your bug fix or feature (if applicable).~~
- [x] Add new passing unit tests to cover the code introduced by your PR (if applicable).
- [] ~~Any breaking changes are specified on the commit on which they are introduced with `BREAKING CHANGE` in the body of the commit.~~
- [] ~~If this is a big feature with breaking changes, consider opening an issue to discuss first. This is completely up to you, but please keep in mind that your PR might not be accepted.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,8 +118,8 @@ The following options can be used when creating an instance of `ImgixClient`:
 - **`path`:** String, required. A full, unencoded path to the image. This includes any additional directory information required to [locate the image](https://docs.imgix.com/setup/serving-images) within a source.
 - **`params`:** Object. Any number of imgix rendering API [parameters](https://docs.imgix.com/apis/url).
 - **`options`:** Object. Any number of modifiers, described below:
-  - [**`disablePathEncoding`**](#disable-path-encoding)
-  - [**`encoder`**](#encoder)
+  - [**`disablePathEncoding`**](#disable-path-encoding): Boolean. Disables encoding logic applied to the image path.
+  - [**`encoder`**](#encoder): Function. Applies custom logic to encode the image path and query parameters.
 
 Construct a single image URL by passing in the image `path` and any rendering API parameters.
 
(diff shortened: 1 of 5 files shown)


atlawrie (approved):
LGTM.

README.md
```diff
@@ -459,6 +459,26 @@ client.buildURL(
...
+
+client.buildURL(
+  "https://proxy.imgix.net/image.jpg",
+  {
+    "txt": "test!(')*"
+  },
+  {
+    encoder: (value, key) => key.substr(-2) === '64' ? Base64.encodeURI(value) : value.replace(' ', "+")
+  }
+)
```
luqven, line 479: I think we should optionally chain here, otherwise, this might throw an error when encoding the path.

```suggestion
    encoder: (value, key) => key?.substr(-2) === '64' ? Base64.encodeURI(value) : value.replace(' ', "+")
  }
)
```

luqven (changes requested):
I think we should add a unit test or two to `_santizePath`. 

Since the `encoder` is used there, that extra optional argument could, and should, throw an error when `key` is not defined if it's not optionally chained/type checked.

README.md
sherwinski: Yep good catch

sherwinski:
At the risk of making this PR lengthier, I patched in some missing TS definitions as well.

test/test-buildURL.js
```diff
@@ -181,6 +182,24 @@ describe('URL Builder:', function describeSuite() {
...
+      const path = 'http://example.com/images and photos/1.png?foo=%20';
+
+      it("encodes the path given custom logic", function testSpec() {
+        const expectation = path.replaceAll(' ', '+');
+        const result = client._sanitizePath(path, { encoder: (path) => path.replaceAll(' ', '+')});
+
+        assert.strictEqual(result.substring(1), expectation);
+      });
+
+      it("does not encode the path any differently if the option second parameter is passed to the custom encoder", function testSpec() {
```
luqven, line 196:
```suggestion
      it("does not alter path encoding when custom encoder optional 'key' parameter is set", function testSpec() {
```

luqven (approved):
LGTM, one non-blocking nit-pick

PR #386, opened 2024-09-19
fix: Upgrade dependencies

icco:
## Description

This does a few things:
 - BREAKING CHANGE: requires node >= 20
 - upgrades rollup to new package schema
 - removes uglifyjs as it's rollup plugin is no longer supported
 - Upgrades packages with known security vulnerabilities
 - Migrates to `.mjs` as required in newer node versions

## Checklist

- [x] Read the [contributing guidelines](CONTRIBUTING.md).
- [x] Each commit follows the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) spec format.
- [x] Update the readme (if applicable).
- [x] Update or add any necessary API documentation (if applicable)
- [x] All existing unit tests are still passing (if applicable).

- [x] Add some [steps](#steps-to-test) so we can test your bug fix or feature (if applicable).
- [x] Add new passing unit tests to cover the code introduced by your PR (if applicable).
- [x] Any breaking changes are specified on the commit on which they are introduced with `BREAKING CHANGE` in the body of the commit.
- [x] If this is a big feature with breaking changes, consider opening an issue to discuss first. This is completely up to you, but please keep in mind that your PR might not be accepted.


marco-salcedo:
@icco thank you so much for contributing! Since this PR is tackling several things we want to make sure we fix your issue without impacting other environments that are using the library.

Could you please provide more insight into what the issue(s) you are running into is/are?
e.g. why do you require dependency updates, need a specific node version, need the files to be imported as `.mjs`, etc

Also, could you please rewrite your commits to make sure we don't check in the distribution files in the PR.

